PR #11, opened 2023-09-25
feat(docs): Add generalized documentation Earthly commands and builders

stevenj:
Generalized Docs Builders/Linters


stevenj:
For review.
The Docs builder seems to work, and has all the plugins we want initially.
The custom Commands for Building, Markdown Linting and repo spellcheck all seem to work fine.
Still need to work out the best way to serve live docs as a development aide.

earthly/docs/Earthfile
```diff
@@ -0,0 +1,175 @@
...
+    ARG tag=latest
+    ARG registry
+
+    # Installation tmp directory
+    WORKDIR /poetry
+
+    # Environment Variables
+    ENV POETRY_HOME=/poetry
+    ENV PATH=$POETRY_HOME/bin:$PATH
+    ENV PYTHONDONTWRITEBYTECODE=1
```
jmgilman, line 18: I see this getting copy/pasted everywhere, and I understand the rationale, but it's also not _always_ required (in the case of this Earthfile, it's a no-op). I'm not opposed to keeping it, but it might be nice to get these into a separate UDC that can be pulled into Python builds. This also allows us to add more documentation around what this actually does.

earthly/docs/Earthfile
```diff
@@ -0,0 +1,175 @@
...
+
+    # Install extra debian packages we will need to support plugins.
+    RUN apt-get update && \
+        apt-get install -y --no-install-recommends \
+        graphviz
+
+    # Install Poetry to manage python.
+    RUN curl -sSL https://install.python-poetry.org | python3 -
+    RUN poetry config installer.max-workers 10
```
jmgilman, line 27: I'd like to know the rationale behind this one. The default is number of cores + 4. So:

- On a Github runner, it's 2 + 4 = 6. 
- On the remote satellite, it's 8 + 4 = 12.
- On the Hetzner machine, it's 32 + 4 = 36.

Notably, the default is _also_ the maximum value you can set. So for example, setting it to 10 on the Github runner is a no-op. For the current remote satellite, we're shaving off 2 cores. Why?

earthly/docs/Earthfile
```diff
@@ -0,0 +1,175 @@
...
+    # Finish building the container, and publish to the registry
+    IF [ "$registry" = "" ]
+        ARG registry_final=$registry
+    ELSE
+        ARG registry_final=${registry}/
+    END
+
+    SAVE ARTIFACT /kroki/docker-compose.yml /kroki/docker-compose.yml
+
+    SAVE IMAGE --push ${registry_final}mkdocs-material:${tag}
```
jmgilman, line 64: Why does this image need to get pushed to a registry? If it's only pulled in locally, I don't see the point. If we do want to actually pull it in remotely, where are we planning on doing that?

earthly/docs/Earthfile
```diff
@@ -0,0 +1,175 @@
...
+    RUN apt-get update && \
+        apt-get install -y --no-install-recommends \
+        graphviz
+
+    # Install Poetry to manage python.
+    RUN curl -sSL https://install.python-poetry.org | python3 -
+    RUN poetry config installer.max-workers 10
+
+    # Copy our dependencies.
+    COPY pyproject.toml .
```
jmgilman, line 30: The `COPY` command can take multiple arguments. I'm not sure if Earthly optimizes for this or not, but technically you've created three layers here when only one is needed (move all of the arguments into a single `COPY` command). If any one of these files change, there's no speed increase by separating them because copy operations are cheap.

earthly/docs/Earthfile
```diff
@@ -0,0 +1,175 @@
...
+    COPY poetry.lock .
+    COPY start.sh .
+
+    # Install it all with poetry
+    RUN poetry install --only main --no-root
```
jmgilman, line 35: The `--no-root` is a no-op here, because there's no root project to begin with. 

Same thing with `-only main` because there's only a single dependency group (`main`). I can see leaving it here for the sake of future compatibility (i.e. not pulling in _future_ dev dependencies).

earthly/docs/Earthfile
```diff
@@ -0,0 +1,175 @@
...
+    CMD ["serve", "--dev-addr=0.0.0.0:8000"]
+
+    # Finish building the container, and publish to the registry
+    IF [ "$registry" = "" ]
+        ARG registry_final=$registry
+    ELSE
+        ARG registry_final=${registry}/
+    END
+
+    SAVE ARTIFACT /kroki/docker-compose.yml /kroki/docker-compose.yml
```
jmgilman, line 62: This seems slightly redundant. The file wasn't modified, why save it as an artifact? Do we intend to mangle it in the future?

earthly/docs/Earthfile
```diff
@@ -0,0 +1,175 @@
...
+
+    FROM earthly/dind:alpine
+    WORKDIR /build
+    COPY +mkdocs-material/kroki/docker-compose.yml .
+
+    # Copy the src
+    COPY --dir $src /docs
+
+    # Docs need Kroki to build diagrams, so make sure its running.
+    WITH DOCKER --compose docker-compose.yml --load mkdocs=+mkdocs-material
```
jmgilman, line 104: As discussed today, try:

```
LOCALLY

RUN docker compose -f docker-compose.kroki.yml
```

IIRC, your `BUILD +mkdocs-material` will have already saved the resulting image locally.

earthly/docs/Earthfile
```diff
@@ -0,0 +1,175 @@
...
+
+    ARG fix
+    ARG src
+
+    RUN --no-cache echo SRC: $src
+
+    # cspell: words davidanson
+    WITH DOCKER --pull davidanson/markdownlint-cli2-rules:next
+        #RUN --no-cache docker run --rm -v $src:/workdir --entrypoint sh davidanson/markdownlint-cli2-rules:next -c "ls -al; ls -al docs"
+        RUN --no-cache docker run --rm -v $src:/workdir davidanson/markdownlint-cli2-rules:next "**/*.md" --config .markdownlint-cli2.jsonc $fix
```
jmgilman, line 121: What is `$fix` doing here? 

This seems really fragile. I'd almost rather vendor the ruleset and then update it as needed. It makes this simpler and also prevents surprises.

earthly/docs/Earthfile
```diff
@@ -0,0 +1,175 @@
...
+    END
+
+CSPELL:
+    # Spell checking all code is done with cspell
+    # See: cspell.org
+    COMMAND
+
+    ARG src
+
+    WITH DOCKER --pull ghcr.io/streetsidesoftware/cspell:latest
```
jmgilman, line 131: I'm a little torn on this approach. It seems DinD can be a bit buggy with Earthly sometimes, and troubleshooting it is not trivial. Are you opposed to just pulling these tools in using their natural patterns instead of running their containers? It's a bit more work, but then you can easily `earthly -i +target` and get into  sane environment to test an issue rather than having to navigate through DinD.

earthly/docs/README.md
```diff
@@ -0,0 +1,11 @@
+# Catalyst SRE Documentation
```
jmgilman, line 1: Probably needs updating :)

earthly/docs/output/kroki/docker-compose.yml
```diff
@@ -0,0 +1,60 @@
+version: "3"
```
jmgilman, line 1: Why is this in version control? I assume it's an output from the Earthly build.

jmgilman (changes requested):

earthly/docs/Earthfile
stevenj: It's not a no-op.  
No problem adding it as a UDC.

The reason I had to do this is because poetry installs everything inside a `venv`,  you can't just run the installed code.  You have to `poetry run` it, or otherwise, get into the `venv` that poetry makes.  And I couldn't find a nice way to do that reliably other than `poetry run`.
But `poetry run` won't work unless you are in the same directory as the `pyproject.toml` that you installed.
So what this does is put poetry into a dependable location `/poetry`, and that's also where the `pyproject.toml` ends up.

I then had to make a shell script to actually run it, so I could still find the `mkdocs.yml` file in the mounted filesystem when the image is used live.

Actually for this image, `poetry` is a PITA.  But it's working.

earthly/docs/Earthfile
stevenj: If you don't, `poetry` throws warning about exceeding the number of workers, when workers exceed 10.  Specifically:

```
Connection pool is full, discarding connection: pypi.org. Connection pool size: 10
```

It throws that error multiple times, and in batches.  Which looks ugly and could make one think it's not working reliably.  I was getting that all the time running locally.

earthly/docs/Earthfile
stevenj: It probably doesn't.

earthly/docs/Earthfile
jmgilman: Sorry, I'm not sure why it highlighted three lines, but I was referencing only line 18:

```
ENV PYTHONDONTWRITEBYTECODE=1
```

AFAIK, since we're just using packages, this _is_ a no-op. Unless there's `.pyc` files getting generated somewhere.

earthly/docs/Earthfile
jmgilman: Wow, that seems really broken 🤣 You would think they would configure the urllib3 pool to match the number of workers. Anyways, for these types of things, I find it helpful to include a comment on why it's being set, especially to help with blind copying/pasting.

earthly/docs/Earthfile
stevenj: `PYTHONDONTWRITEBYTECODE=1` stops the python interpreter from saving pointless bytecode cache files inside the short-running container, which will only be lost when the container stops anyway.
It's a small perf improvement, python has the bytecode in memory anyway.

An improvement on this would be to pre-interpret all the python into bytecode and actually save it inside the container.  But that seems unnecessary.

earthly/docs/Earthfile
jmgilman: https://github.com/python-poetry/poetry/issues/8469

earthly/docs/Earthfile
stevenj: This could be my misunderstanding.  I thought one needed to `SAVE ARTIFACT` to `COPY` it into another target?

earthly/docs/Earthfile
stevenj: So mdlint can autofix certain violations.  Especially handy when you don;t have the right file spacing and what not.
So `$fix` was my hacky way of using one rule to do two things.  `lint only`  or `lint and fix`.
On a developers machine, they probably always want to run `lint and fix` in CI we only want `lint`.
I will see if I can think of a cleaner way.

earthly/docs/Earthfile
jmgilman: Nevermind, I didn't understand what you were doing when I first made this comment. I see the reason this is needed since it's being pulled into a UDC (where `COPY` wouldn't work the way we need it).

earthly/docs/Earthfile
jmgilman: It saves some lines to remove this and all the registry boiler plate. I would just leave the `SAVE ARTIFACT`.

earthly/docs/Earthfile
jmgilman: Yeah, I get why it makes sense for a long-running container (i.e. FastAPI), but for containers that essentially get thrown away (i.e. build and output the docs), I was thinking most of the benefits were negated. 

But yes, this is a good candidate for going into a UDT. Not sure what it would look like, but probably some sort of "Python setup" UDT that optimized the compiler in these sorts of ways. I'll see about tracking it in a separate issue.

earthly/docs/Earthfile
stevenj: I need the image to run the mkdocs builder in `serve` mode.
It doesn't need to be pushed to a repo.

earthly/docs/Earthfile
stevenj: I have reworked all this implementation to do it a different way and not try and run doc serving inside earthly.

earthly/docs/Earthfile
stevenj: I do need the $fix option, but I made it so it can only accept the one allowed value (or blank).  Anything else will fail the build.

earthly/docs/output/kroki/docker-compose.yml
stevenj: Yes, I culled it.  Accidentally got checked in.

earthly/docs/Earthfile
stevenj: Its a bit of work to get it right, and then we have to maintain it.
If we think its necessary, I would rather that be the subject of an alternative PR.
For now, I have moved both linters to their own earthfiles, so this kind of refactoring is easier.
I would advocate we leave it for now, as I have bigger fish to fry, and we revisit it if it causes breakage in CI.

stevenj:
This is a substantial update to the previous PR, and is ready for re-review.

earthly/cspell/Earthfile
```diff
@@ -0,0 +1,38 @@
...
+    # See: cspell.org
+    COMMAND
+
+    # Where we want to run the `lint` from.
+    ARG --required src
+
+    # Unlikely to need to change this.
+    ARG cfg_file=cspell.json
+
+    RUN --no-cache if [[ ! -f $src/$cfg_file ]]; then \
```
jmgilman, line 15: I don't think the `--no-cache` flags are required here. Think of it like this:

```
COPY src src
DO +CSPELL --src=src
```

In the above case, if the contents of the `src` directory change, then the cache breaks at that layer and the `CSPELL` UDC will always be run. If the contents of `src` _do not_ change, then normally the `CSPELL` UDC wouldn't run. However, the explicit `--no-cache` flags may override that behavior and cause this UDC to always run. 

The net effect is that doing this may _always_ bust the cache when this UDC is used and cause downstream layers to constantly rebuild.

earthly/cspell/Earthfile
```diff
@@ -0,0 +1,38 @@
...
+        (>&2 echo "$src/$cfg_file does not exist. Can not check spelling."); \
+        exit 1; \
+    fi
+
+    RUN --no-cache echo Spell Checking Recursively from: $src
+
+    WITH DOCKER --pull ghcr.io/streetsidesoftware/cspell:latest
```
jmgilman, line 22: I still think it's worth considering installing this in its native package form rather than using `WITH DOCKER`.

earthly/cspell/Readme.md
```diff
@@ -0,0 +1,78 @@
...
+multiple source languages and documentation files.
+Tool specific spell checkers are limited in the kinds of files they can check,
+and also will use different configurations, dictionaries and project word lists.
+
+## Invocation
+
+In an Earthfile in your repo, add the following:
+
+```earthfile
+spellcheck-lint:
```
jmgilman, line 24: Is there a particular reason we promote running it locally? I'm not sure I understand the justification at the end. The typical flow is copy source files into the container and then lint/check/test/build. In some cases, especially when doing code generation, the source files may change and be different than the ones stored locally.

earthly/cspell/Readme.md
```diff
@@ -0,0 +1,78 @@
...
+    LOCALLY
+
+    DO github.com/input-output-hk/catalyst-ci:v1.2.0/earthly/cspell+CSPELL --src=$(echo ${PWD})
+```
+
+In this use case, the UDC is run Locally, so that the src in the repo can be directly checked.
+
+## Configuration
+
+Each repo will need a [`cspell.json`](http://cspell.org/configuration/) file in the root of the repo.
```
jmgilman, line 35: Github offers "template repositories" which might make this process a bit smoother. Or we can use something like cookie cutter to generate template repos.

earthly/python/Earthfile
```diff
@@ -0,0 +1,34 @@
...
+FROM python:3.11-bullseye
+
+# Common Poetry setup
+# Parameters:
+#   opts      : Options passed to the `poetry install` command.
+#   extra_file: An extra file to copy into the poetry path.
+POETRY_SETUP:
+    COMMAND
+    ARG opts
+    ARG extra_file
```
jmgilman, line 12: I think we can make this `extra_files` and have it be a space-separated list of filenames.

earthly/docs/Earthfile
```diff
@@ -0,0 +1,92 @@
+VERSION 0.7
+FROM python:3.11-bullseye
+
+mkdocs-material:
+    # Derived from official mkdocs-material docker container.
+    # https://github.com/squidfunk/mkdocs-material/blob/master/Dockerfile
+
+    # Install poetry and our python dependencies.
+    DO ../python+POETRY_SETUP --extra_file=start.sh
```
jmgilman, line 9: Passing `start.sh` as an extra file is a bit misleading here as it doesn't actually do anything to change the `poetry install` invocation. I think it's a bit more clear to just have it be its own `COPY` line.

earthly/docs/Readme.md
```diff
@@ -0,0 +1,11 @@
+# Catalyst Documentation Builders and UDCs
+
+This directory contains targets for the Catalyst Documentation builders, and associated UDCs.
+
+## Build
```
jmgilman, line 5: This is out of date.

earthly/docs/services/Readme.md
```diff
@@ -0,0 +1,19 @@
+# Documentation Service Definitions
+
+<!-- cspell: words livedocs sitedocs -->
+
+## Live Documentation
+
+To build documentation live, from the root of the repo run:
+
+```sh
+docker compose up -f local/docker-compose.livedocs.yml up --abort-on-container-exit
```
jmgilman, line 10: Just a thought: in the past we'd typically use a Nix devshell in order to add a "task runner" to our repos. In other words, special commands that could be easily remembered and invoked to perform common operations. This isn't unique to Nix, though. We can start using an actual task runner to encapsulate some of these wordy invocations.

earthly/docs/services/docker-compose.sitedocs.yml
```diff
@@ -0,0 +1,12 @@
+version: "3"
```
jmgilman, line 1: I'm wondering if the distinction is really necessary here? In what use case would someone like to view the docs without live reloading? You can emulate this by using `serve` and just _not_ touching the files. 

I say this as I already had this exact thought and chose not to go with it for the above reason.

earthly/mdlint/Earthfile
```diff
@@ -0,0 +1,61 @@
...
+
+    # Status line for what we are about to do.
+    RUN --no-cache echo Linting Markdown Recursively from: $src
+
+    # Run the linter with the given arguments, and recursively check all markdown files.
+    # The directory to be checked `MUST` have a `.markdownlint-sli2.jsonc` file.
+    # cspell: words davidanson
+    WITH DOCKER --pull davidanson/markdownlint-cli2-rules:next
+        # Lint checks should not be cached.
+        RUN --no-cache \
```
jmgilman, line 43: See previous comment about using `--no-cache`

earthly/mdlint/Earthfile
```diff
@@ -0,0 +1,61 @@
...
+        exit 1; \
+    fi
+
+    # Status line for what we are about to do.
+    RUN --no-cache echo Linting Markdown Recursively from: $src
+
+    # Run the linter with the given arguments, and recursively check all markdown files.
+    # The directory to be checked `MUST` have a `.markdownlint-sli2.jsonc` file.
+    # cspell: words davidanson
+    WITH DOCKER --pull davidanson/markdownlint-cli2-rules:next
```
jmgilman, line 41: See previous comment about using `WITH DOCKER`.

earthly/mdlint/Earthfile
```diff
@@ -0,0 +1,61 @@
...
+            $fix
+    END
+
+# A Test and example invocation of the above UDC.
+mdlint-test:
+    # Test Markdown lint checks.
+    # Run with `earthly -P +mdlint-test
+    LOCALLY
+
+    ARG src=$(echo ${PWD}/../../)
```
jmgilman, line 59: Does `${PWD}` resolve to the location of the Earthfile or to the location that `earthly` was invoked?

earthly/mdlint/Readme.md
```diff
@@ -0,0 +1,47 @@
...
+This is because we need to provide uniform and consistent Markdown formatting and linting across the project and between projects.
+
+## Invocation
+
+In an Earthfile in your repo, add the following:
+
+```earthfile
+markdown-lint:
+    # Check Markdown in this repo.
+    LOCALLY
```
jmgilman, line 21: See previous comment about recommending LOCALLY here.

jmgilman (changes requested):

earthly/cspell/Earthfile
saibatizoku: I think that this is a good way to enable swapping the cspell image, which I assume is already optimized and loaded with the configured binary.

earthly/python/Readme.md
```diff
@@ -0,0 +1,35 @@
...
+Once the UDC has run, the Earthly target that invoked it will
+
+#### Invocation
+
+In an `Earthfile` in your source repository add:
+
+```Earthfile
+example_python_target:
+    FROM python:3.11-bullseye
+    DO DO github.com/input-output-hk/catalyst-ci:v1.2.0/earthly/python+POETRY_SETUP
```
saibatizoku, line 19: Typo: double `DO`.

earthly/cspell/Earthfile
stevenj: There are two contexts here.
1. Its run in CI on the Checked out code of a PR, before any containers start being built.
2. Its run on the local machine, checking the developer's code in their repo.
In both cases it's running on code which is not inside any container being built.
So Earthly has no way to know if the code changed or not.  Better for it to always run the test than falsely say nothing changed.

stevenj: There is no layer.

earthly/cspell/Earthfile
stevenj: I do not believe there is any justifiable return on investment in re-constructing this already working container.  Its out of scope for this PR.

earthly/cspell/Readme.md
stevenj: Because its running on Local code in the developers repo, or a fresh checkout in CI before any containers start getting built.  Its not running as part of a container being built.  And in fact the majority of time someone would want to run this the last thing they would want is to actually build a container.  They just want to know, do I have any markdown lint issues in the code I am working on.

Once its been run on the src, there is zero point in running it on code that's then been copied into a container.

earthly/cspell/Readme.md
stevenj: We will have 5 Repos total this applies to.
We need to be really conscious about spending time to automate things that are trivial in practice and happen once.
I am sure I can copy this config file into those 5 repos faster than we can automate it.

earthly/python/Earthfile
stevenj: I tried that and it didn't work at all.  The argument was taken as a single file named: "file1 file2 file3".
It is my preference that it work that way, so if you can suggest how to make it work, more than happy to add it.

earthly/docs/services/docker-compose.sitedocs.yml
stevenj: Its 100% necessary.
I already have examples where live docs do not look the same as built docs.
Some plugins generate files and references to them very poorly.  Like they use relative paths in live docs, or worse, paths relative to the URL of the live server.
And in built docs, its a whole other path relative to the supposed URL you will deploy it on.  or other ways.  There doesn't;t seem to be any consistency.
Being able to check both will ensure I make docs which not only look ok to me when I write them, but won't be a broken mess when we deploy them.
This is a critical feature for me.
Finding some of the reasons why this happens is also on my TODO list, but its much lower importance to a lot of other things, so initially just knowing what I wrote is broken means I can quickly work around it and make working docs.

earthly/mdlint/Earthfile
stevenj: Previous reply also applies here. Its to be run locally, not inside a container being built.

earthly/mdlint/Earthfile
stevenj: Same reply, ultra-low priority with no clear return on investment vs the cost of reimplementing something that already works.  Especially given the ongoing maintenance needed to track upstream changes as and when they occur.

Mdlint, and CSpell are not just a single file that's run and done. They package with plugins and data files and what not.  All of which we would be taking on the burden of maintaining.  Until I see this actually cause an issue with a reproducable example I can't see any value in it.

I also don't see any reason it can't work OK if `FROM` can happily pull a Container published to a container repo?

We could go:

```Earthly
my-mdlint:
  FROM davidanson/markdownlint-cli2-rules:next
```

And that's no different from pulling alpine or debian.  The big difference here is we don;t want to customize it at all.

earthly/mdlint/Earthfile
stevenj: My testing indicates it's the location of the Earthfile the target is in.
I use this kind of thing because earthly refuses to accept that `../../` is a valid path but `./` is.  It needs an absolute path if its not a child path of the Earthfile for some odd reason.

earthly/mdlint/Readme.md
stevenj: Same reply.  its not running inside a container build.

earthly/python/Readme.md
stevenj: Fixed.

earthly/cspell/Earthfile
stevenj: Am I misunderstanding what your saying here?
Its not clear if your suggesting we re-make the cspell image, or we do this in an alternative way?

earthly/mdlint/Earthfile
stevenj: Again,  is it your suggestion to do this an alternative way, or to rebuild the markdownlint container ourselves?

earthly/mdlint/Earthfile
stevenj: This is one of my Earthly/Docker bugbears.  Any path that's valid should be valid.  If I can have a path relative to the Earthfile, I should be able to step up to a parent.  If its a security issue allow it with a CLI option.

stevenj:
I tagged this PR as https://github.com/input-output-hk/catalyst-ci/releases/tag/t1.2.2 so that I could test integration with docs being built in the `catalyst-internal-docs`.  That flushed out a few issues, but Markdown and Spelling lint fine.  The docs build OK and can be served Live and as a static site locally for full local testing/development.

earthly/docs/services/docker-compose.sitedocs.yml
jmgilman: Interesting, I wrongly assumed they would be the same 😅 You would think that `serve` would hot-compile the same static site that `build` does...either way, makes sense to have both until we can figure out what's going on.

earthly/cspell/Earthfile
jmgilman: Sure, but this is a result of doing it locally instead of copying the files into the container. If I call this UDC in `TargetA` and then include that target in `TargetB`, then `TargetB` will never get cached because of the usage of `--no-cache` here. 

It seems like not using `LOCALLY` solves this issue.

earthly/cspell/Earthfile
jmgilman: I'm just looking to standardize the way we include external tools. Historically, for something like `cspell`, I would have done:

```
FROM node:20-bookworm
RUN npm install -g git+https://github.com/streetsidesoftware/cspell-cli
RUN cspell lint .
```

I've typically avoided `WITH DOCKER` because of some prior experiences where it was:

1. Slow
2. Didn't seem to cache the way I would have expected (i.e. I'm not sure if DinD caches the image)
3. Involved more effort to figure out how to user the Docker image correctly rather than just run the binary itself

In this case, the second one doesn't apply because we're technically using the local workstation as a cache (with `LOCALLY`), but that also sort of breaks the "isolation" that Earthly is providing (i.e. we're now creating a side effect). 

Whatever we want to do, I just want to make sure we do it across all Earthfiles for the sake of consistency.

earthly/cspell/Readme.md
jmgilman: Understandable. I suppose I'm just gunshy to the idea of breaking isolation by using `LOCALLY`. The Earthly documentation also suggests caution when using this:

```
The LOCALLY command can be used in place of a FROM command, which will cause earthly to execute all commands under the target directly on the host system, rather than inside a container. Commands within a LOCALLY target will never be cached. This feature should be used with caution as locally run commands have no guarantee they will behave the same on different systems.
```

Both of those problems are valid: we're breaking isolation and caching at the same time - two things that Earthly works to guarantee against. 

Theoretically, I don't think there's a speed increase either. A user still has to wait for the Docker image to download to their machine before the command can execute. Versus pulling the source into a container and executing it there, which would still have a delay for the container to build,  but it would only happen once and then be cached. So both methods have some sort of first-time delay and the difference is probably negligible to really consider it a factor.

earthly/python/Earthfile
jmgilman: That's too bad, probably some shell escaping going on. We can leave it for now and come back to it later if it really is a problem.

earthly/cspell/Earthfile
stevenj: Why would we be linting inside the container?  We shouldn't even build the container if the lints fail?
The whole point of this is to check things are good before we do more complex stuff, like compile code.

stevenj: It seems torturous also to force that to do any listing we must build a container just to hold the code we want to lint.
All I want to do is check does the code on my disk pass before I commit it to the repo, or that I just checked out during a PR action pass.

Its not something we would ever use when making a container.

earthly/cspell/Earthfile
minikin: @jmgilman 
Thank you for your insights. Maintaining multiple third-party packages like mdlint and cspell may be challenging, given our limited resources. Finding a balance between standardization and maintenance overhead is crucial to ensure we remain focused on our core project deliverables. Your suggestion for consistency is well-noted, and finding a simplified, low-maintenance approach to include necessary external tools is a priority.

earthly/cspell/Readme.md
minikin: @stevenj @jmgilman 
We should implement a spell checker that requires minimal engineering effort to run locally.

earthly/cspell/Earthfile
jmgilman: Let's give it a shot. It would be helpful to get a working example with all the pieces in place. Then we can do some tests, especially around how things interact with `--no-cache`.

earthly/cspell/Earthfile
jmgilman: Sure, I also only want to maintain what we need to. Let's go with it and we can always come back to this in the future if we start running into issues.

jmgilman (approved):

PR #12, opened 2023-09-27
feat: adds typescript based actions

jmgilman:
This PR overhauls the existing Github Actions to consolidate some and convert a few to be Typescript based. This allows greater flexibility (over bash) and enables easier testing on actions.


stevenj (approved):
Nice

PR #46, opened 2023-10-17
feat: Rust builder setup

stevenj:
Ready to go.
I have a test release tagged against this, which I am using to test the Earthly Rust build in Cat Voices.
As soon as this merges, I will delete it, re-tag an official release with this and update the cat-voices rust earthly to point to the correct ci version.

### DIFF
--- a/earthly/rust/Earthfile
+++ b/earthly/rust/Earthfile
@@ -55,7 +55,6 @@ rustup:
 RUST_SETUP:
     COMMAND
     ARG toolchain=./rust-toolchain.toml
-    ARG host=x86_64-unknown-linux-musl
 
     # Poetry Installation directory.
     # Gives poetry and our poetry project a reliable location.
@@ -67,9 +66,8 @@ RUST_SETUP:
 
     # Install pinned Rustup from `rust-toolchain.toml`
     # Plus nightly latest so we can use it for docs, lints, etc.
-    RUN rustup show && \
-        rustup set default-host $host && \
-        rustup default $default_rust_channel && \
+    RUN rustup default $default_rust_channel && \
+        rustup show && \
         rustup toolchain install nightly --component miri --component rust-src && \
         cargo --version && \
         cargo +nightly --version
--- a/earthly/rust/Earthfile
+++ b/earthly/rust/Earthfile
@@ -54,7 +54,7 @@ rustup:
 #  * toolchain : The `rust-toolchain` toml file.
 RUST_SETUP:
     COMMAND
-    ARG toolchain
+    ARG toolchain=./rust-toolchain.toml
 
     # Poetry Installation directory.
     # Gives poetry and our poetry project a reliable location.
--- a/earthly/rust/Earthfile
+++ b/earthly/rust/Earthfile
@@ -1,0 +1,73 @@
+# Common Rust UDCs and Builders.
+VERSION 0.7
+
+# Base Rustup build container.
+rustup:
+    # Base it on Debian bookworm-slim
+    FROM debian:bookworm-slim
+
+    WORKDIR /root
+
+    # Install necessary packages 
+    # Expand this list as needed, rather than adding more tools in later containers.
+    RUN apt-get update && \
+        apt-get install --no-install-recommends -y \
+            ca-certificates \
+            curl \
+            file \
+            build-essential \
+            autoconf \
+            automake \
+            autotools-dev \
+            libtool \
+            xutils-dev \
+            mold \
+            python3 \
+            python3-pip \
+            musl-tools && \
+        rm -rf /var/lib/apt/lists/*
+
+    # Install OpenSSL here but ONLY if we can't avoid using it.
+
+    # Install rustup.
+    RUN curl https://sh.rustup.rs -sSf | sh -s -- --default-toolchain none -y -v
+
+    # Set necessary env vars from this setup.
+    ENV PATH=/root/.cargo/bin:$PATH
+
+# Common Rust setup
+# Parameters:
+#  * toolchain : The `rust-toolchain` toml file.
+RUST_SETUP:
+    COMMAND
+    ARG toolchain
+
+    # Poetry Installation directory.
+    # Gives poetry and our poetry project a reliable location.
+    WORKDIR /build
+
+    # Copy our toolchain dependency.
+    COPY $toolchain ./rust-toolchain.toml
+    ENV default_rust_channel=$(grep -oP 'channel\s*=\s*"\K[^"]+' rust-toolchain.toml)
+
+    # Install pinned Rustup from `rust-toolchain.toml`
+    # Plus nightly latest so we can use it for docs, lints, etc.
+    RUN rustup show && \
+        rustup default $default_rust_channel && \
+        rustup toolchain install nightly --component miri --component rust-src && \
+        cargo --version && \
+        cargo +nightly --version
+
+    # Install tools we use commonly with `cargo`.
+    RUN cargo install cargo-nextest --locked && \
+        cargo install cargo-chef --locked && \
+        cargo install kani-verifier --locked && \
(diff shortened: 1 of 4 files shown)


earthly/rust/Earthfile
```diff
@@ -0,0 +1,75 @@
...
+            curl \
+            file \
+            build-essential \
+            autoconf \
+            automake \
+            autotools-dev \
+            libtool \
+            xutils-dev \
+            mold \
+            python3 \
```
jmgilman, line 27: Why do we need Python?

earthly/rust/Earthfile
```diff
@@ -0,0 +1,75 @@
...
+            python3-pip \
+            musl-tools && \
+        rm -rf /var/lib/apt/lists/*
+
+    # Install OpenSSL here but ONLY if we can't avoid using it.
+
+    # Install rustup.
+    RUN curl https://sh.rustup.rs -sSf | sh -s -- --default-toolchain none -y -v
```
jmgilman, line 35: This isn't a deal breaker, it is more of an FYI. These `curl | bash` install methods are pretty risky inside of a build environment. Think of it from the role of an attacker: they can clearly see we use `rustup` and are using the default install method. If they manage to take ownership of the HTTP response (legitimately or illegitimately), we're effectively running an unknown payload with root permissions. 

This might seem innocent, given this is a builder that gets thrown away...but it's not far-fetched to infect a build environment and start maliciously manipulating builds. 

Again, not trying to be pedantic, but if there's an alternative install method, or at the very least a signature we can use to validate the response...that would be nice.

earthly/rust/Earthfile
```diff
@@ -0,0 +1,75 @@
...
+
+    # Set necessary env vars from this setup.
+    ENV PATH=/root/.cargo/bin:$PATH
+
+# Common Rust setup
+# Parameters:
+#  * toolchain : The `rust-toolchain` toml file.
+RUST_SETUP:
+    COMMAND
+    ARG toolchain
```
jmgilman, line 45: You could probably set a default value here, since the file is typically always given the same name.

earthly/rust/Earthfile
```diff
@@ -0,0 +1,75 @@
...
+
+    # Poetry Installation directory.
+    # Gives poetry and our poetry project a reliable location.
+    WORKDIR /build
+
+    # Copy our toolchain dependency.
+    COPY $toolchain ./rust-toolchain.toml
```
jmgilman, line 52: I think you _might_ be duplicating work. IIRC, the `rustup` script will install a default version of Rust as the initial toolchain. I think there's a way to have it instead installed from a toolchain file.  It's a slight optimization, as it avoids installing Rust twice.

earthly/rust/Earthfile
```diff
@@ -0,0 +1,75 @@
...
+    # Install tools we use commonly with `cargo`.
+    RUN cargo install cargo-nextest --locked && \
+        cargo install cargo-chef --locked && \
+        cargo install kani-verifier --locked && \
+        cargo kani setup && \
+        cargo install refinery_cli --locked
+
+
+# Test rust build container
+rust-builder:
```
jmgilman, line 72: You should either rename this target to `check` or add a `check` target that builds this one. Then CI will grab it and include it as part of the checks to let us know if someone breaks this.

jmgilman (changes requested):

earthly/rust/Earthfile
stevenj: The `kani-verifier` needs it as a build dependency.
Rather than scatter `apt-get install` all over the place I figured its best just to do all the dependencies in one place.

earthly/rust/Earthfile
stevenj: I agree, but they don't have a way I can find to validate it beyond relying on https to validate the domain.  Which I have read replies to this topic suggesting that should be enough.
The official rust docker image does this, it only adds a SHA256 hash check, which they hard code.
We can't do that easily because we want to align the rust version with the `rust-toolchain.toml` of the project being built.
And I can't find a way to get an authoritative list of the hashes, and if I could we would still be back to relying on https to ensure the hash list is accurate, so its a catch-22.
The best would be if they signed it, and we knew the public key.  But they don't.

The easiest supply chain hack on rust would be to hack their docker github repo, point it to a malicious version, and re-release a malware package.  It looks like they use github CI to sign and publish, so that would still sign and publish the malware.

earthly/rust/Earthfile
stevenj: There isn't.  I went through that.  I specifically don't install a default toolchain when I install rustup, just rustup itself.
Ideally we wouldn't be changing the base os + rustup build very often once we have a mature build process.


So the idea was, just install `rustup` in the base os image and no rust version at all.
Then install the rust version defined by the `rust-toolchain.toml` itself.
Thats why I used `--default-toolchain none` on the rustup install.

The way rustup works is it installs the latest rust by default, or a specific named version, ie `1.72` or `stable` or `nightly`, etc.
`stable` always installs the latest version, which we may not need or use, especially if the base-os is a couple of rust releases behind (which is possible), so the latest rust when the `rustup` layer was built in CI, may not be the latest version by the time we use it to do a build.  So installing one then seems redundant.  It also doesn't mark the version it installs from `rust-toolchain.toml` as a default and there was no way I could find to get it to do so.  

I think the intention is `rustup` installs many SYSTEM versions of rust.  One can be manually set to the SYSTEM default, but if a `rust-toolchain.toml` is present, it will use that version and install it if asked, but it doesn't;t make it default.  I had to jump through hoops to get it to make the `rust-toolchain.toml` actually be the default by extracting the version from it and manually telling `rustup` to do it.  its not really designed to be ergonomic for a container build use case at all.

`rustup` has no facility to tell it to "use" a particular `rust-toolchain.toml` it just picks it up if its present in the CWD.  And if you said `rustup install stable` it ignores it completely and installs the latest stable rust.  Which isn't what we want.

All said `rustup` is a PITA, but I have worked around its shortcomings.

earthly/rust/Earthfile
jmgilman: > We can't do that easily because we want to align the rust version with the rust-toolchain.toml of the project being built.

Not sure I'm following. The script itself likely remains fairly static. What's preventing us from doing the same thing: hashing a known version and embedding it in the Earthfile. It's a pain to update, but like I said, I wouldn't imagine it changing _that_ much.

earthly/rust/Earthfile
stevenj: no problem :)
It is intended as a test that the container build process hangs together.

earthly/rust/Earthfile
stevenj: Well, then why not just put `rustup.sh` in the `/earthly/rust/` directory and copy it locally.  No need to download or hash check at all.
Debian has added `rustup` package, but its unstable.  Otherwise we could install it like anything else:
https://ftp-master.debian.org/new.html

stevenj: I Just triple-checked.  It doesn't do any validation.  100% relies on https/tls1.3 to be secure to ensure it's getting the correct files.  So, it's no less secure than what we do here.  Changing it just adds maintenance work with no increase in security.  It doesn't even hash check anything.

jmgilman: Yeah, not really surprising. Maybe I'll file an upstream issue if there isn't already one.

earthly/rust/Earthfile
jmgilman: Yeah, I had a similar experience, but it gets the job done I guess.

earthly/rust/Earthfile
stevenj: Done

jmgilman (approved):
LGTM

PR #52, opened 2023-10-24
feat(docs): Add docs check target, plus some extra plugins

stevenj:
Adding docs features as I stand up Catalcy Vopices docs.


earthly/docs/Earthfile
```diff
@@ -48,39 +49,50 @@ MKDOCS_BUILD:
     COPY +mkdocs-material/services/docker-compose.kroki.yml .
 
     # Copy the src
+    RUN mkdir /extras 
     COPY --dir $src /docs
+    FOR --sep=", " file IN $extra_files
```
jmgilman, line 54: We should standardize on space-separated for inputs that need multiple values. This is already the standard in other places, and it's also the default for bash (i.e., you can remove the `--sep` flag).

docs/Earthfile
```diff
@@ -1,16 +1,35 @@
...
+
+# Make a locally runable container that can serve the docs.
+local:
+    # Build a self contained service to show built docs locally.
+    DO ../earthly/docs+PACKAGE
+
+    # Copy the static pages into the container
+    COPY +docs/ /usr/share/nginx/html
+
+    # This is a local only image, we do not publish it.
```
jmgilman, line 34: Just FYI, we don't use `--publish` at all anymore, because the CI handles the publishing. The comment seems to suggest otherwise.

docs/Earthfile
```diff
@@ -1,16 +1,35 @@
...
 
-    COPY --dir src mkdocs.yml .
+    # Just copy the actual doc src here.
+    COPY --dir src includes macros overrides mkdocs.yml /docs
 
+    # Now copy into that any artifacts we pull from the builds.
+    COPY --dir ../+repo-docs/repo /docs/includes
+
+
+# Build the docs here.
```
jmgilman, line 16: Nitpick: can we come to an agreement on the level of verbosity that we want to hit with comments? This target reads like a narrative, but I'm not sure it's really necessary. I try to save comments for tricky parts that are not obvious by just looking at it. 

You can see my other Earthfiles. Most of them have no comments except for edge cases that need clarity. So there's a pretty dramatic difference between our two styles :)

docs/mkdocs.yml
```diff
@@ -1,10 +1,120 @@
-site_name: Catalyst CI Documentation
+# cspell: words Diagramsnet, pymdownx, linenums, inlinehilite, superfences,
+# cspell: words tasklist, betterem, materialx, twemoji, smartsymbols,
+# cspell: words kroki, excalidraw, glightbox, cips, pygments, arithmatex
+
+# Project Information
+site_name: Project Catalyst - Catalyst CI Documentation
```
jmgilman, line 6: On my end, this looks pretty ugly. The name is too long and wraps in the navigation bar. Maybe something in the middle, like: "Project Catalyst CI Docs" ?

docs/src/guides/languages/flutter.md
jmgilman, line 1: For these, can we:

1. Capitalize the title
2. Make a small note about the content coming soon? Otherwise, it's just a confusing blank page, and someone might think their browser is broken.

docs/mkdocs.yml
```diff
@@ -1,10 +1,120 @@
...
+      media: "(prefers-color-scheme: dark)"
+      toggle:
+        icon: material/brightness-4
+        name: Switch to light mode
+
   features:
-    - navigation.indexes
+    - navigation.instant
+    - navigation.instant.progress
     - navigation.tracking
```
jmgilman, line 39: I'm not sure which setting is doing this, but the navigation sidebar now includes sections, instead of showing a ToC on the left side. I think I really preferred it the other way, because some documents have a lot of headings, and it completely blows up the navigation bar and makes it really hard to figure out where you are.

CONTRIBUTING.md
jmgilman, line 1: There are a lot of links in here that still point to catalyst-voices.

docs/mkdocs.yml
```diff
@@ -21,13 +135,17 @@ nav:
...
-  - pymdownx.tabbed:
-      alternate_style: true
+      - Documentation Examples:
+          - Examples: appendix/examples/index.md
+          - Abbreviations: appendix/examples/abbreviations.md
+          - Admonitions: appendix/examples/admonitions.md
+          - Code Formatting: appendix/examples/code-formatting.md
+          - Mermaid Diagrams: appendix/examples/mermaid-diagrams.md
+          - Kroki Diagrams: appendix/examples/kroki-diagrams.md
+      - Important Information:
```
jmgilman, line 145: How did you get icons in this section? It looks kind of weird that the documentation examples and important information both have icons, but nothing else does.

jmgilman (changes requested):

earthly/docs/dev/local.py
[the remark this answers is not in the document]
stevenj: Ok,  what this is, is a script to allow me (or anyone) to do "live docs" editing.
It uses the fact earthly caches so good, to allow it to periodically re-run the doc build locally, and then locally re-deploy the container with the built docs serving on `8123`.
I haven't documented how to use it yet and that's on the todo list.
I am going to be making a bunch of tickets for the devs to each pick one up and write the "how to use" section for each of the main targets, using this would be part of `docs` docs.

docs/Earthfile
stevenj: I am simply trying to highlight that this package is NOT a deployment package, it's for local use only.  It's also why I used `local` and not one of the normal targets.

docs/Earthfile
stevenj: My `comment` knob was stuck at `11` many years ago.  I fear if I try and force it back down I will just break it off instead.

What we could do, is leave them in for now,  I am going to be making a ticket for a dev to add the "how to use docs" part of the cat-ci guide documentation.  At that time, they can eliminate the comments I have here, and translate it into the guide page itself.

docs/mkdocs.yml
stevenj: No problem.

docs/mkdocs.yml
stevenj: Will check it.
You are right to nitpick this because I am wanting the cat-ci mkdocs.yml to be our preferred starting point for all docs.

docs/mkdocs.yml
stevenj: I think the icons really enhance the look of the docs.
It's not hard to add them.
I will fix the TOC issue you mentioned above first, and then look at adding icons to at least the top level pages.
The icon is added using a bit on yml frontmatter to the markdown, or inside a special file which sets yml frontmatter metadata for all files beside and below it.

docs/Earthfile
```diff
@@ -1,16 +1,35 @@
...
 docs:
     FROM +src
 
-    RUN poetry run mkdocs build --strict
-    SAVE ARTIFACT site/*
\ No newline at end of file
+    DO ../earthly/docs+BUILD
+
+    RUN /scripts/build.sh
```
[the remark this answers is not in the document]
stevenj: Ok, just checked this, its a spillover from an edit.  The ONLY part of the build should be the `DO`.  Everything it does is standard and no need to customize it in any way.
All the customization goes into `src:`.

docs/mkdocs.yml
jmgilman: Yeah, I think adding icons to the others would be fine. It just looks funny when some have it and some don't.

docs/mkdocs.yml
stevenj: New navigation style applied for review.

docs/mkdocs.yml
stevenj: Done

docs/mkdocs.yml
stevenj: Done

docs/Earthfile
stevenj: Done

docs/Earthfile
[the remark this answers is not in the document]
stevenj: Done

Issue #53, opened 2023-10-24
Fix rust build for macOS m1


Mr-Leshiy:
Need to fix rust build for macOS m1 platform. 
Right now `earthly ./earthly/rust+check` fails with the following errors:
```
cannot produce proc-macro for `async-trait v0.1.74` as the target `aarch64-unknown-linux-gnu` does not support these crate types
```

PR #54, opened 2023-10-24
Updated aarch64-unknown-linux-gnu rustflags, updated rust Earthly targets

Mr-Leshiy:
Close https://github.com/input-output-hk/catalyst-ci/issues/53

- fixed aarch64-unknown-linux-gnu rustflags, to solve an issue with cargo-nextest build, disable `target-feature=+crt-static` flag
- updated rust Earthly targets, added `with_cargo_nextest`, `with_cargo_chef`, `with_kani`, `with_refinery` and `with_cargo_machete` earthly arguuments, so it is possible to conditionally install these tools.
- fixed following issues on the `aarch64-unknown-linux-gnu` when eathly rust targets are runned on the macOS M1:
```
cannot produce proc-macro for `async-trait v0.1.74` as the target `aarch64-unknown-linux-gnu` does not support these crate types
```
```
Error: Kani does not support this platform (Rust target aarch64-unknown-linux-gnu)
```

### DIFF
--- a/earthly/rust/Earthfile
+++ b/earthly/rust/Earthfile
@@ -52,20 +52,9 @@ rustup:
 # Common Rust setup
 # Parameters:
 #  * toolchain : The `rust-toolchain` toml file.
-#  * with_cargo_nextest: <true/false> Flag to install cargo-nextest tool
-#  * with_cargo_chef: <true/false> Flag to install cargo-chef tool
-#  * with_kani: <true/false> Flag to install kani-verifier and kani tools
-#  * with_refinery: <true/false> Flag to install refinery tool
-#  * with_cargo_machete: <true/false> Flag to install cargo-machete tool
 RUST_SETUP:
     COMMAND
     ARG toolchain=./rust-toolchain.toml
-
-    ARG with_cargo_nextest=false
-    ARG with_cargo_chef=false
-    ARG with_kani=false
-    ARG with_refinery=false
-    ARG with_cargo_machete=false
 
     # Poetry Installation directory.
     # Gives poetry and our poetry project a reliable location.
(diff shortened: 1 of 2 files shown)


earthly/rust/Earthfile
```diff
@@ -52,10 +52,21 @@ rustup:
...
 RUST_SETUP:
     COMMAND
     ARG toolchain=./rust-toolchain.toml
 
+    ARG with_cargo_nextest=false
+    ARG with_cargo_chef=false
+    ARG with_kani=false
+    ARG with_refinery=false
+    ARG with_cargo_machete=false
+
```
stevenj, line 69: No, we are not having optional tooling.

earthly/rust/Earthfile
```diff
@@ -73,18 +84,40 @@ RUST_SETUP:
...
+    IF [ "$with_kani" = "true" ]
+        RUN cargo install kani-verifier --locked && \
+        cargo kani setup
+    END
+    IF [ "$with_refinery" = "true" ]
+        RUN cargo install refinery_cli --locked
+    END
+    IF [ "$with_cargo_machete" = "true" ]
+        RUN cargo install cargo-machete --locked
+    END
```
stevenj, line 102: See above.  A tool either works everywhere or we don;t use it.

earthly/rust/config.toml
```diff
@@ -13,14 +13,14 @@ rustflags = [
...
-    # "-C", "target-feature=+crt-static" - Doesn't work for nextest.
+    # "-C", "target-feature=+crt-static" - proc-macro doesn't work with it. `https://github.com/rust-lang/rust/issues/78210`
 ]
 
 [target.aarch64-unknown-linux-gnu]
 linker = "clang"
 rustflags = [
     "-C", "link-arg=-fuse-ld=/usr/bin/mold",
-    "-C", "target-feature=+crt-static"
+    # "-C", "target-feature=+crt-static" - proc-macro doesn't work with it. `https://github.com/rust-lang/rust/issues/78210`
```
stevenj, line 23: Thats not unexpected.  The key point is to make sure when your target is built its NOT built with this target.  This target can ONLY be used to build the compiler and its tooling.

earthly/rust/config.toml
```diff
@@ -13,14 +13,14 @@ rustflags = [
...
 ]
 
 [target.aarch64-unknown-linux-musl]
```
stevenj, line 26: This is the target catalyst-gateway should build with on an apple mac (or any AArch64 based device actually).

stevenj (changes requested):

earthly/rust/Earthfile
stevenj: You can eliminate `kani` and `chef`.
* `chef` just breaks builds, it was on my TODO to eliminate it anyway.
* `kani` either works everywhere or we don;t use it.  Just comment it our with a note saying. "can not be used because not supported on Mac M1+ devices"  So that next time someone suggests to use it we can see why we don't.

* refinery, nextest and machete should work everywhere and are not optional.

earthly/rust/config.toml
Mr-Leshiy: But it does not work for nextest, the same issue as with `x86_64-unknown-linux-gnu`

stevenj: nextest is just a tool, its ok to build it with the GNU tools.  I hit this with x86 build as well.  This is OK to disable for getting nextest to run.

stevenj (approved):

PR #75, opened 2023-11-02
fix: Add -P flag for earhtly `test` CI runner

Mr-Leshiy:
It is needed to pass `-P` flag to earthly during tests running to fix the following issue https://github.com/input-output-hk/catalyst-ci/actions/runs/6733556300/job/18302659923

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -125,6 +125,7 @@ jobs:
       ci_cli_version: ${{ inputs.ci_cli_version }}
       earthly_version: ${{ inputs.earthly_version }}
       target: test
+      privileged: true
     secrets:
       dockerhub_token: ${{ secrets.dockerhub_token }}
       dockerhub_username: ${{ secrets.dockerhub_username }}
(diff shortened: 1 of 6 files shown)


jmgilman:
Let's modify this so that the `test` target is always given the `-P` flag. This will require changing a few things:

1. The [run.yml](https://github.com/input-output-hk/catalyst-ci/blob/master/.github/workflows/run.yml) workflow will need an input added for passing earthly flags. This input needs to then be fed into the [flags input for the run action.](https://github.com/input-output-hk/catalyst-ci/blob/master/actions/run/action.yml#L15). 
2. The [ci.yml](https://github.com/input-output-hk/catalyst-ci/blob/master/.github/workflows/ci.yml) workflow will then need to be updated to pass the `-P` flag to the new input added above [during the test step](https://github.com/input-output-hk/catalyst-ci/blob/master/.github/workflows/ci.yml#L119).

Mr-Leshiy:
@jmgilman seems it is ready all check passed. You can check for this commit https://github.com/input-output-hk/catalyst-ci/pull/75/commits/3036b963e48c0d60f0f2bdff607f590226f9449f.

jmgilman (approved):
LGTM. Thanks for the help!

PR #92, opened 2023-11-14
feat(docs): PostgreSQL guide

### LINKED ISSUE #80: Add Guide for SQL
SQL needs a guide similar to the GO Guide.

Mr-Leshiy:
# Description

Added PostgreSQL builder guide.

## Related Issue(s)

Closes #80


Mr-Leshiy:
deployed docs: https://input-output-hk.github.io/catalyst-ci/branch/feat_sql_guides/guides/languages/postgresql/

docs/src/guides/languages/postgresql.md
```diff
@@ -9,4 +9,301 @@ tags:
...
+```
+
+There are 4 possible options how to run:
+
+* If `DB_HOST` env var established to `localhost`,
+PostgreSQL server runs as a part of the `example` service,
+otherwise will relies on remote PostgreSQL server connection
+(as an example already defined `postgres` service).
+* `INIT_AND_DROP_DB` env var defines to run initial initialization of the db with the clean state or not
+(could be omitted).
```
stevenj, line 179: Use the word `optional` instead.  But also put a description of what happens when the option is NOT specified.  Does it default, does it cause the container to behave differently? 

Same comment applies to each option below.

stevenj (changes requested):

docs/src/guides/languages/postgresql.md
Mr-Leshiy: fixed

stevenj (approved):
LGTM

PR #99, opened 2023-11-20
feat(docs): Python guide

### LINKED ISSUE #79: Add Guide for Python
Python needs a guide similar to the GO Guide.

Mr-Leshiy:
# Description

- Added python guide section to docs
- Added python example project
- Updated python builder

## Related Issue(s)

Closes #79 


earthly/python/Earthfile
```diff
@@ -26,8 +29,16 @@ POETRY_SETUP:
...
+# Parameters:
+#   opts      : Options passed to the `poetry install` command.
+BUILDER:
+    COMMAND
+    ARG opts
+
     # Copy our dependencies.
-    COPY pyproject.toml poetry.lock $extra_file .
+    COPY pyproject.toml .
+    COPY --if-exists poetry.lock .
```
stevenj, line 41: `poetry lock` MUST exist.  If it doesn't the python project has not properly vendored its dependencies.
Also, why did you remove the ability to copy an extra file here?

docs/src/guides/languages/python.md
```diff
@@ -9,4 +9,89 @@ tags:
...
+
+    COPY --dir ./src .
+    DO ./../../earthly/python+BUILDER
+```
+
+The first target `builder` is responsible to prepare an already configured Python environment,
+instal all needed tools and dependencies.
+Every Python project must be a [poetry](https://python-poetry.org) based project,
+so it is mandatory to have `pyproject.toml` file in the root dir of the project.
+
```
stevenj, line 61: Also mandatory to vendor the dependencies. `poetry.lock` needs to exist and be up-to-date.

stevenj (changes requested):
See comments, plus even though cat-ci does not yet enforce it, the title needs to be a conventional commit PR title.

earthly/python/Earthfile
Mr-Leshiy: To have a generalized builder I have followed these suggestions https://python-poetry.org/docs/basic-usage/#as-a-library-developer. 
But if you think that we should include this file every time and we are not going to have python libs, I will reverse this change.

Mr-Leshiy: >why did you remove the ability to copy an extra file here?

I have removed it from this place because it will be much easier to COPY needed files outside the UDC command, 
as it was done here https://github.com/input-output-hk/catalyst-ci/blob/48502f89eed3af79af5ed3641f92b9f35c89674e/examples/python/Earthfile#L6 . UDC arguments could not provide so wide functionality in the convenient way (like passing some options to `COPY` command)

stevenj: Yes, we should enforce a vendored and uptodate lock file for now.  If it turns out later we don't need it we can worry about that then.  I don't plan on us making python libraries anytime soon.  We only use python for Automation, testing and "glue" tasks, all of which are executables.

stevenj: I am ok with the removal of extra file copy.

docs/src/guides/languages/python.md
Mr-Leshiy: fixed

earthly/python/Earthfile
Mr-Leshiy: updated

stevenj (approved):
LGTM

PR #100, opened 2023-11-21
refactor(docs): Move common docs into a common sub-dir to relieve directory clutter.

stevenj:
Refactors docs so that common documentation is in a `common` sub directory.
This helps keep the intention of the files clear and reduces directory clutter in the `earthly/docs` directory.
Also re-introduces a `common` target to hold the common docs, as they are not a dependency of the builder itself.
It is reasonable if a documentation target does not wish to use the standard them, and therefore does not need the `common` files but does need the documentation builder.


Mr-Leshiy (approved):
LGTM

PR #107, opened 2023-11-24
feat(docs): Rust guide

### LINKED ISSUE #78: Add Guide for Rust
Rust needs a guide in the CI documentation, similar to the GO Guide.

Mr-Leshiy:
# Description

- Added Rust project building guide.
- Refactored Rust builder, moved example Rust project to the `examples/rust` dir.
- Made some cleanup

## Related Issue(s)

Closes #78

* [ ] My code follows the style guidelines of this project
* [ ] I have performed a self-review of my code
* [ ] My changes generate no new warnings
* [ ] New and existing unit tests pass locally with my changes


Mr-Leshiy:
Docs link: https://input-output-hk.github.io/catalyst-ci/branch/feat_rust_guide/

stevenj (changes requested):
Sorry, you have made a ton of unrelated changes to the code outside of the scope for the ticket.
The ticket did not ask you to rename +BUILDER to +SETUP that is disturbing to anyone else working on the code base, and also breaks cat-voices.
Even if that's a change we want to make and I don't know why we would, that needs to be a ticket all its own, and each target should be addressed separately.
Please revert all changes that have nothing to do with Rust documentation.

The task also was not to refactor or re-write the rust builder. So again, revert all that.
If you think the rust builder needs correction it needs to be raised in a ticket and properly discussed.

I have things in the builder for reasons.  Especially to help us debug if things go wrong.
It doesn't;t cost us anything to print out the build platform data at the start of setting up the container.

Your setup assumes that rust-toolchain is with the Earthfile, it may not be.
There are so many changes to the rust builder that I can;t reason through it.
Your task was not to re-write it.
It was to document it ONLY.

I will not approve this PR unless it has only the following:
1. Rust documentation.
2. Fixes to demonstratable issues discovered while documenting it.
3. You CAN move the rust example to the examples folder, but it should just be a move, and the changes in it should  be limited to just that move (paths changing for exxample).

earthly/docs/Readme.md
stevenj, line 1: Needs to be reverted not a subject of this ticket

earthly/rust/Earthfile
```diff
@@ -84,16 +80,13 @@ rust-base:
...
 # Parameters:
 #  * toolchain : The `rust-toolchain` toml file.
 SETUP:
     COMMAND
-    ARG toolchain=./rust-toolchain.toml
 
-    # Poetry Installation directory.
-    # Gives poetry and our poetry project a reliable location.
-    WORKDIR /build
+    ARG toolchain=./rust-toolchain.toml
```
stevenj, line 89: I think its better for UDC arguments to appear after the COMMAND so they are clear when looking at it.
I shouldn't have to read the whole UDC to find its arguments.

earthly/rust/Earthfile
```diff
@@ -110,22 +103,6 @@ SETUP:
...
-    ARG src=""
-
-    # ONLY copy whats needed to build.  Not everything.
-    # Note: rust-toolchain.toml was already copied when rust was setup.
-    # Minimizing whats copied reduces needless cache misses.
-    FOR --sep=", " file IN $src
-        COPY --dir $file .
-    END
-
-
```
stevenj, line 128: Use in cat-gateway and so removing this breaks that.
Do not remove it.

docs/src/guides/languages/rust.md
```diff
@@ -9,4 +9,211 @@ tags:
...
+### Prepare base builder
+
+```Earthfile
+VERSION 0.7
+
+# Set up our target toolchains, and copy our files.
+builder:
+    FROM ./../../earthly/rust+rust-base
+    
+    COPY --dir "*" .
```
stevenj, line 58: We specifically don't want this.  the reason CP_SRC exists is to make it easy to ONLY copy relevant source files.
We should not be copying files which are not part of the rust build.
For example cat-gateway has the full event-db schema in it.  None of that has anything to do with rust, copying it would cause the target to rebuild when the schema changes, but the rust code itself did not.
Or worse, if we add seed data.  None of which should trigger a rebuild of the rust target.

earthly/rust/Earthfile
```diff
@@ -135,114 +112,20 @@ CHECK:
...
     RUN target/$TARGETARCH/release/$bin --help
-
-
-# Set up our target toolchains, and copy our files.
-builder:
-    FROM +rust-base
-
-    DO +SETUP --toolchain=example/rust-toolchain.toml
-
-    DO +CP_SRC --src="example/*"
```
stevenj, line 158: This needs to be in the example, but make it more like cat-gateway now that its moved.

examples/rust/Earthfile
```diff
@@ -0,0 +1,79 @@
+VERSION 0.7
+
+# cspell: words TARGETARCH USERARCH toolsets
+
+# Set up our target toolchains, and copy our files.
+builder:
+    FROM ./../../earthly/rust+rust-base
+
+    COPY --dir "*" .
```
stevenj, line 9: Use CP_SRC and only copy files that affect the rust build itself. Exclude anything that does not, such as Readme.md
or any other files not part of the rust build.

docs/src/guides/languages/rust.md
Mr-Leshiy: I agree with your point to copy only needed files.
But `CP_SRC` actually does not solve this problem, it acts as the same as a regular `COPY` command but with less functionality which `COPY` has.

stevenj (approved):
OK

PR #120, opened 2023-12-05
feat(docs): Add Rust dependency graph generation, enable workspace indexing for Rust docs

Mr-Leshiy:
# Description

- Added Rust dependency graph generation
- enabled workspace indexing for Rust docs
- Added Rust module graph generation
- Extended current Rust example project with more than one crate
- Updated Rust guide

## Related Issue(s)

Related to https://github.com/input-output-hk/catalyst-voices/issues/148


earthly/rust/scripts/std_build.sh
```diff
@@ -18,6 +18,8 @@ source "$(dirname "$0")/colors.sh"
 
 rc=0
 
+export RUSTDOCFLAGS="--enable-index-page -Zunstable-options"
```
stevenj, line 21: Shouldn't this be defined in `stdcfgs/cargo_config.toml` and then in the projects `config.toml` to match?

earthly/rust/scripts/std_build.sh
```diff
@@ -42,6 +44,11 @@ status $rc "Checking Documentation tests all pass" \
 status $rc "Checking Benchmarks all run to completion" \
     cargo bench --all-targets; rc=$?
 
+## Generate dependency graphs
+cargo depgraph --workspace-only --dedup-transitive-deps > target/doc/workspace.dot
+cargo depgraph --dedup-transitive-deps > target/doc/full.dot
+cargo depgraph --all-deps --dedup-transitive-deps > target/doc/all.dot
+
 ## Generate Module Trees for documentation purposes.
 # cargo modules generate tree --orphans --types --traits --fns --tests --all-features --lib
```
stevenj, line 53: Would be good to enable these as well and add them to the docs, as they will show us the internal structure of our modules.
We can generate a graph, as well as the tree.  I think both are useful.
For the `--bin` target we will need to know the name of the binary.  We could pass it as a parameter to std-build.sh and get it in the Earthfile `BUILD` UDC.  We would then need to supply the binary name.  If no binary name was supplied we could skip generation for `--bin` and only generate for --lib.

earthly/rust/scripts/std_build.sh
Mr-Leshiy: fixed

earthly/rust/scripts/std_build.sh
Mr-Leshiy: done

earthly/rust/scripts/std_build.sh
Mr-Leshiy: I have figured out that, added these options into the `cargo_config.toml` file. I was the problem with the `cargo-modules` version, the latest one does not work

FelipeRosa (approved):
LGTM

PR #128, opened 2023-12-11
feat: scanning all targets related to the given `target`

### LINKED ISSUE #111: [Feature]: Allow targets to be found by a wildcard.
### Feature Title

Multiple targets in a single Earthfile

### Feature Description

Currently we have `check`, `build`, `test`, etc targets, 
There are some cases where we need multiple `check` or other targets in the same Earthfile.
We handle that now by running `BUILD` on multiple targets from the main `check` or whatever target.
The problem is they don't run in parallel, it increases maintenance and the first that fails stops other `checks` or whatever target from executing.

The feature request here is to extend what targets will get recognized.
* `check` + `check-*`
* `build` + `build-*`
* `package` + `package-*`
* `test` + `test-*`
* `release` + `release-*`
* `publish` + `publish-*`

### Use Case

Multiple targets are defined in a single Earthfile.
Quite useful for `check` and also `test` targets, but can be generally applied to all targets to give us better flexibility.

### Is this a modification to an existing feature?

Yes

bkioshn:
# Description
Scanning all targets in the following pattern `target` or `target-*`

## Related Issue(s)

>Closes  #111 

## Description of Changes

-  Fix the `ScanForTarget` to return a map where key is the Earthfile path and value is a struct of `Earthfile` and a list of filtered targets that suite the input `target
2 Patterns can be executed here
    - `target` - should match the exact word `target`
    - `target-*` - the `target` with hyphen followed by * should match target that start with `target` with hyphen and followed by numbers or characters 
**:rotating_light: DO NOT USE WILDCARD with target that produce artifact or image**
- `EarthTargets` struct is added to `earthfile.go` to support the map mentioned above
- `get_target` step is added to `run.yml` `publish.yml` `release.yml` to extract the targets that needed to be run by getting it from the discover stage.
- Remove `check` in `Earthfile` since `check-*` can be used
- Use `@actions/exec` instead of `exec` in `child_process`
- image flag is now removed from `discover` stage 
- output in `discover` stage now ouput `json` a map of Earthfile and filtered targets and a path of Earthfile needed to be run 
- local flag is included in `install/action.yml` in order to build the Cat-CI CLI locally instead of using from a specific github version
- in `run` workflow, filtered target that matched that input `target` will be run in different process

- Test cases for target scanning including
    - Target exact match
    - Target with wildcard (*)
    - Mismatch target
- Test cases for `discover` workflow
    - add check for expected `json` and `path` output
- Test cases for `install` workflow
    - test for `local` flag and non `local` flag
- Test cases for `run` workflow 
    - spawn command should be equal to the command length

## Breaking Changes
- Change the branch before merge
- Change the local flag to `false`
- Apply the new flow to `release` and `publish`
- Related docs need to be updated once the new flow is discussed


cli/pkg/scanners/file_scanner.go
```diff
@@ -29,9 +30,15 @@ func (f *FileScanner) Scan() ([]pkg.Earthfile, error) {
...
+	// Should start with given target.
+	// Allowing an optional hyphen followed by one or more lowercase letters or numbers
+	regexPattern := "^" + regexp.QuoteMeta(target) + "(?:-[a-z0-9]+)?$"
+
+	r, _ := regexp.Compile(regexPattern)
+
 	earthfiles, err := f.scan(func(e pkg.Earthfile) (bool, error) {
 		for _, t := range e.Targets {
-			if t.Name == target {
+			if r.MatchString(t.Name) {
```
jmgilman, line 41: So the issue here is this will change the current behavior of the CLI. As it stands, passing `-t target` will _only_ return targets with the exact name of `target`. With this change, the CLI will _also_ return `target-*` targets. 

I typically try to avoid making backward incompatible changes unless they are really necessary. It's frustrating for a flag to suddenly change its logic fairly dramatically. Instead, we should force the end-user to explicitly specify when they intend to glob targets. For example:

```
-t target -t target-*
```

Would match `target` and `target-test`.

```
-t target
```

Would _only_ match `target` and nothing else.

jmgilman (changes requested):

.github/workflows/ci.yml
```diff
@@ -95,7 +95,7 @@ jobs:
       aws_region: ${{ inputs.aws_region }}
       ci_cli_version: ${{ inputs.ci_cli_version }}
       earthly_version: ${{ inputs.earthly_version }}
-      target: check
+      target: check-*
```
Mr-Leshiy, line 98: I think it would be better to specify a target name just as `check`, `build`, `test` etc, without any special characters at this place.

.github/workflows/ci.yml
```diff
@@ -109,7 +109,7 @@ jobs:
       aws_region: ${{ inputs.aws_region }}
       ci_cli_version: ${{ inputs.ci_cli_version }}
       earthly_version: ${{ inputs.earthly_version }}
-      target: build
+      target: build-*
```
Mr-Leshiy, line 112: same

.github/workflows/ci.yml
```diff
@@ -137,7 +137,7 @@ jobs:
       aws_region: ${{ inputs.aws_region }}
       ci_cli_version: ${{ inputs.ci_cli_version }}
       earthly_version: ${{ inputs.earthly_version }}
-      target: test
+      target: test-*
```
Mr-Leshiy, line 140: same

.github/workflows/ci.yml
jmgilman: No, we don't want this, as it breaks backward compatibility. By specifying a `-*` we are indicating we want to capture more than a single target. We don't want to enforce capturing multiple targets with no way to go back to the original form.

actions/install/action.yml
```diff
@@ -9,6 +9,10 @@ inputs:
     description: The version of the Catalyst-CI CLI to install
     required: false
     default: latest
+  local:
+    description: Use the local Catalyst-CI CLI
```
jmgilman, line 13: We should indicate in the description that this is _building_ a local version of the CLI. Otherwise, it might be confusing about where it's coming from.

actions/install/src/install.ts
```diff
@@ -1,6 +1,7 @@
 import * as core from '@actions/core'
 import * as tc from '@actions/tool-cache'
 import * as github from '@actions/github'
+import { exec } from 'child_process'
```
jmgilman, line 4: Let's change this out for GitHub's toolkit `exec` package: https://github.com/actions/toolkit/tree/main/packages/exec. It has better cross-platform support and provides async support. It's also easier to mock.

actions/run/src/run.ts
```diff
@@ -103,3 +115,24 @@ async function spawnCommand(command: string, args: string[]): Promise<string> {
     })
   })
 }
+
+function getTargetsFromEarthfile(target: string, earthfile: string): string[] {
```
jmgilman, line 119: Perhaps I'm missing something here, but why are we parsing the targets in Typescript? Isn't this something we're already doing with the CLI?

actions/run/src/run.ts
```diff
@@ -33,30 +35,40 @@ export async function run(): Promise<void> {
     args.push(...flags.split(' '))
   }
 
-  if (artifact) {
-    args.push('--artifact', `${earthfile}+${target}/`, `${artifactPath}`)
-  } else {
-    args.push(`${earthfile}+${target}`)
-  }
+  const targets = getTargetsFromEarthfile(target, earthfile)
```
jmgilman, line 38: This should come from executing the CLI. It's probably a good time to make an improvement here and change the `target` input to be a list of targets instead of a single target. The CLI already supports passing multiple targets by specifying the `-t` flag multiple times.

jmgilman (changes requested):

actions/run/src/run.ts
bkioshn: Yes this is what I concern about, it is quite redundant to `getTarget` in `earthfile.go`, but isn't the `scan` in `cli/cmd/main.go` only set the output for the path that contains earthfile?  Correct me if I'm wrong

.github/workflows/ci.yml
Mr-Leshiy: Got it, I see, thanks for explanation

cli/pkg/scanners/file_scanner.go
```diff
@@ -28,21 +31,57 @@ func (f *FileScanner) Scan() ([]pkg.Earthfile, error) {
...
 
 	if err != nil {
 		return nil, err
 	}
-	return earthfiles, nil
+
+	return pathToEarthTargets, nil
+}
+
+func getTargetRegex(target string) string {
```
Mr-Leshiy, line 75: Pls add some docs about this function, why it is needed and how it works

cli/cmd/main.go
```diff
@@ -74,91 +72,71 @@ func (c *imagesCmd) Run() error {
 
 type scanCmd struct {
 	JSONOutput bool     `short:"j" long:"json"   help:"Output in JSON format"`
-	Images     bool     `short:"i" long:"images" help:"Also output images for the target of each Earthfile (requires -t option)"`
 	Paths      []string `                        help:"paths to scan for Earthfiles"                                             arg:"" type:"path"`
-	Target     string   `short:"t"               help:"filter by Earthfiles that include this target"                                               default:""`
+	Target     []string `short:"t"               help:"filter by Earthfiles that include this target"                                               default:""`
```
Mr-Leshiy, line 76: why do we need array here ?

Mr-Leshiy, line 76:
```suggestion
	Target     []string `short:"t"               help:"filter by Earthfiles that include these targets"                                               default:""`
```

bkioshn: Yeah that's true, should be string

bkioshn: My bad, it should be array since we can pass `-t target -t target2`

actions/run/action.yml
```diff
@@ -29,12 +29,13 @@ inputs:
     description: The port to use for connecting to the remote runner
     required: false
     default: "8372"
-  target:
-    description: The name of the target to run
+  targets:
+    description: The name of the targets to run
```
jmgilman, line 33: Change to, "A space separated list of targets to run"

actions/run/dist/index.js
```diff
@@ -2901,26 +2902,40 @@ async function run() {
...
-        core.info(`Found image: ${imageOutput}`);
-        core.setOutput('image', imageOutput);
-    }
-    const artifactOutput = external_path_.join(earthfile, parseArtifact(output));
-    if (artifactOutput !== earthfile) {
-        core.info(`Found artifact: ${artifactOutput}`);
-        core.setOutput('artifact', artifactOutput);
+    core.info(`Filtered targets >> ${targets}`);
+    const targetsArr = targets.split(' ');
+    for (const tg of targetsArr) {
```
jmgilman, line 2910: This seems a bit verbose. Why not just split the targets and then map the result? We don't need to log anything here either.

actions/run/dist/index.js
```diff
@@ -2901,26 +2902,40 @@ async function run() {
...
+        // Get the filtered targets associated with the pattern target and earthfile.
+        core.info(`Pushing target ${tg}`);
+        targetsArgs.push(`${earthfile}+${tg}`);
+    }
+    // Running each target command in different process.
+    for (const t of targetsArgs) {
+        core.info(`Target: ${t}`);
```
jmgilman, line 2917: Change this to: `Running target: ${t}`

actions/run/dist/index.js
```diff
@@ -2901,26 +2902,40 @@ async function run() {
...
+            argsSpawn.push(t);
+        }
+        core.info(`Running command: ${command} ${argsSpawn.join(' ')}`);
+        const output = await spawnCommand(command, argsSpawn);
+        const imageOutput = parseImage(output);
+        if (imageOutput) {
+            core.info(`Found image: ${imageOutput}`);
+            core.setOutput('image', imageOutput);
+        }
+        const artifactOutput = external_path_.join(earthfile, parseArtifact(output));
```
jmgilman, line 2934: So this will be a bit of a bug. In the original design, there was only ever one target, which means only a single artifact is ever produced. Now, if two targets are run, the `artifact` output will be set to the last one that ran (which is probably confusing). 

Fixing this will be a bit of work. If we want, for now, we just need to specify in the documentation that it doesn't support handling multiple targets that each produce an artifact.

The same issue exists with images.

.github/workflows/ci.yml
```diff
@@ -188,7 +188,7 @@ jobs:
       earthly_runner_address: ${{ secrets.earthly_runner_address }}
       earthly_runner_secret: ${{ secrets.earthly_runner_secret }}
   deploy:
-    uses: input-output-hk/catalyst-ci/.github/workflows/deploy.yml@master
+    uses: input-output-hk/catalyst-ci/.github/workflows/deploy.yml@feat/targets-scanner
```
jmgilman, line 191: Make sure to change all of these back before merging.

actions/discover/src/discover.ts
```diff
@@ -1,21 +1,36 @@
...
+    for (const key in parsedData) {
+      if (Object.prototype.hasOwnProperty.call(parsedData, key)) {
+        pathsArray.push(key)
+      }
+    }
+
+    // JSON of path mapping to its filtered list of targets that need to be run
+    core.setOutput('json', stdout)
+    // List of path that should be run
+    core.setOutput('paths', pathsArray)
```
jmgilman, line 33: Can you help me understand why we need two distinct outputs here? It looks like `paths` is just a transformed version of `json`.

.github/workflows/run.yml
```diff
@@ -1,102 +1,113 @@
...
       fail-fast: false
       matrix:
-        earthfile: ${{ fromJson(needs.discover.outputs.json) }}
+        earthfile: ${{ fromJson(needs.discover.outputs.paths) }}
     steps:
+      - name: Get filtered targets
+        id: get_target
+        run: |
+          targets=$(echo '${{ needs.discover.outputs.json }}' | jq -r --arg key '${{ matrix.earthfile }}' '.[$key][]')
+          echo $targets
```
jmgilman, line 105: We can keep this here, but we should make it a bit cleaner. Something like: `echo "Found targets: $targets"`

actions/install/action.yml
```diff
@@ -9,6 +9,11 @@ inputs:
     description: The version of the Catalyst-CI CLI to install
     required: false
     default: latest
+  local:
+    description: Build and use the local version of the Catalyst-CI CLI
+    required: false
+    # TODO - Change default
```
jmgilman, line 15: Reminder to not forget this.

cli/cmd/main.go
```diff
@@ -74,91 +72,71 @@ func (c *imagesCmd) Run() error {
 
 type scanCmd struct {
 	JSONOutput bool     `short:"j" long:"json"   help:"Output in JSON format"`
-	Images     bool     `short:"i" long:"images" help:"Also output images for the target of each Earthfile (requires -t option)"`
 	Paths      []string `                        help:"paths to scan for Earthfiles"                                             arg:"" type:"path"`
-	Target     string   `short:"t"               help:"filter by Earthfiles that include this target"                                               default:""`
+	Targets    []string `short:"t"               help:"filter by Earthfiles that include this targets pattern"                                               default:""`
```
jmgilman, line 76: "this targets pattern" -> "these target patterns"

docs/src/guides/languages/rust.md
```diff
@@ -81,15 +81,15 @@ By default `toolchain` setup to `rust-toolchain.toml`.
 
 ```Earthfile
 # Test rust build container - Use best architecture host tools.
-check-hosted:
+hosted-check:
```
jmgilman, line 84: Why did these all get reversed?

docs/src/reference/actions.md
```diff
@@ -41,9 +41,10 @@ Using these actions individually should be avoided unless absolutely necessary.
 ### Discover
 
 The `discover` action is another common action that shows up in many workflows.
-It performs the "discovery" mechanism of finding Earthfiles with specific targets.
-For example, the `check` workflow uses this action to discover all Earthfiles that have a `check` target.
-The custom CI CLI **must** be installed (see the above section) in order for this action to work.
+It performs the "discovery" mechanism of finding Earthfiles with filtered targets from specific targets.
+For example, the check workflow, which run `check` and `check-*`,
+uses this action to discover all Earthfiles that match `check` and `check-*` targets.
+The custom CI CLI **must** be installed or locally build (see the above section) in order for this action to work.
```
jmgilman, line 47: locally build -> built locally

jmgilman (changes requested):

actions/discover/src/discover.ts
bkioshn: so it is easirt to manage in `run.yml`
```
 - name: Discover Earthly files
        uses: input-output-hk/catalyst-ci/actions/discover@feat/targets-scanner
        id: discover
        with:
          targets: ${{ inputs.target }}
      - name: Check for empty output
        id: check
        run: |
          json=$(echo '${{ steps.discover.outputs.json }}' | jq -rc)
          paths=$(echo '${{ steps.discover.outputs.paths }}' | jq -rc)
          if [ "$output" == "null" ]; then
            echo "json=[]" >> $GITHUB_OUTPUT
            echo "paths=[]" >> $GITHUB_OUTPUT
          else
            echo "json=$json" >> $GITHUB_OUTPUT
            echo "paths=$paths" >> $GITHUB_OUTPUT
          fi
```

actions/run/dist/index.js
bkioshn: Maybe just set it to all targets? Not sure how should it be.

docs/src/guides/languages/rust.md
bkioshn: we want this check to be done locally. Already discuss with @stevenj

actions/discover/src/discover.ts
bkioshn: To avoid calling the filtering target multiple times, I decided to retrieve the map of Earthfile paths and their corresponding filtered targets. Consequently, the output here needs to be modified from the previous version. Instead of `json` being a list of Earthfile paths, this list is now placed in `paths`. The `json` now represents all the discovered Earthfiles along with a list of filtered targets. This information will be used to extract the necessary targets to run during the matrix.

actions/run/dist/index.js
bkioshn: Current Resolution: for now wildcard should only be implemented to targets that don't produce any artifact or image. This will be mentioned in the document.

Mr-Leshiy (approved):
LGTM

Mr-Leshiy (approved):

jmgilman (approved):
LGTM

PR #129, opened 2023-12-12
feat: add a docs target into the pipeline

### LINKED ISSUE #112: [Feature]: Add a documentation target
### Feature Title

`docs:` target to run at the end of builds.

### Feature Description

We currently have a specific `doc` target.
It will need artifacts from other targets.
It should be triggered at the end of a CI run, not automatically on every PR so that it can take advantage of the build caches from the targets optimally.

The change here is to trigger docs build `pages.yml` during a CI run, at the same time that the "release" and "publish" steps would run.
We also need to remove the current docs trigger `docs.yml` which currently triggers this.

Need to test that docs properly still build for both `main`/`master` branches and also for other branches.
This ticket includes updating both catalyst-voices and hermes to use this new CI flow, and removing their `docs.yml` github actions, and testing those repos still build docs correctly. 

Don't make multiple tickets for this, just link each PR to this issue.

### Use Case

Improving cache hit and performance of building docs.

### Is this a modification to an existing feature?

Yes

apskhem:
# Description

Add a documentation target.

## Related Issue(s)

> e.g., Closes #112 

## Description of Changes

* Added `docs` target to trigger `pages.yml`
* Removed `docs.yml`
* Updated the onboarding docs adding the new `docs` stage


stevenj:
This looks like what we need, but I notice that this PR isn't running any actions now, except for the semantic pull request.    Do you know why?

Mr-Leshiy (approved):
LGTM

PR #139, opened 2023-12-29
feat: implement catalyst ci simulate command

### LINKED ISSUE #130: 🛠️ [TASK] :Simulate the pipeline using command line
### Summary

Simulate the pipeline using command line

### Description

Simulate the pipeline using command line.
The current pipeline can be seen [here](https://input-output-hk.github.io/catalyst-ci/onboarding/).

`simulate` command
- Running Earthly targets according to the given input `target` 
- Each target should be run sequentially, no need to run in parallel.
Simulated CI run should stop on the first error.

`generate` command
- Generate a Earthfile containing a list of given input `target`
- This Earthfile can then be called by `earthly` command
 
both of the command take a list of targets to run through. 
If not defined, it should default to `check -> build -> test`

bkioshn:
# Description
Simulate the pipeline using command line.
The current pipeline can be seen [here](https://input-output-hk.github.io/catalyst-ci/onboarding/).

## Related Issue(s)

> Closes #130 

## Description of Changes
- Earthly target can be run simultaneously locally
- Argument "path" is required
- Argument "target" is list of target patterns that wanted to run 
   - If not defined the default targets will be run [check check-* build test test-*]
   - eg. `ci simulate . -t "test test-*"`
 

## Related Pull Requests

> #128 


cli/cmd/main.go
```diff
@@ -227,6 +236,123 @@ func (c *tagsCmd) Run() error {
...
+	Path    string   `                      help:"directory path to be iterated to search for targets within the Earthfile"               arg:"" type:"path"`
+	Targets []string `short:"t"               help:"Earthly targets pattern"                                               default:"check check-* build test test-*"`
+}
+
+func (c *simulateCmd) Run() error {
+	parser := parsers.NewEarthlyParser()
+	scanner := scanners.NewFileScanner([]string{c.Path}, parser, afero.NewOsFs())
+
+	fmt.Println("Targets: ", c.Targets)
+	// Loop through target patterns.
```
stevenj, line 249: I would rather we turn this into a separate function, which just collects a sorted array of targets.
The code here then would call that, and then iterate the returned array to run each target.

cli/cmd/main.go
```diff
@@ -227,6 +236,123 @@ func (c *tagsCmd) Run() error {
...
+	// Write down Earthly version and main target.
+	setup := fmt.Sprintf("VERSION --global-cache %s \nsimulate:\n", c.Version)
+	_, err = writer.WriteString(setup)
+	if err != nil {
+		return nil
+	}
+
+	parser := parsers.NewEarthlyParser()
+	scanner := scanners.NewFileScanner([]string{c.Path}, parser, afero.NewOsFs())
+	// Loop through target patterns.
```
stevenj, line 329: Same as above.  We really should have a function which gets the sorted list of targets, and then we should iterate that to create the earthfile.  In this case, and the case above the function to get the target list should be the same.

stevenj (changes requested):

PR #159, opened 2024-01-17
chore: Add catalyst-pallas (pallas lib fork repo) to the allowed sources list in deny.toml

FelipeRosa:


Mr-Leshiy (dismissed):
LGTM

earthly/rust/stdcfgs/deny.toml
```diff
@@ -268,7 +268,7 @@ allow-git = []
 
 [sources.allow-org]
 # 1 or more github.com organizations to allow git sources for
-#github = [""]
+github = ["input-output-hk"]
```
stevenj, line 271: I would prefer we don't do this, for the pallas fork especially.
The pallas fork should not be long lived, we should be upstreaming any changes we do ASAP.
If we DO use a github repo directly, I would also rather we specifically declare it and not a whole organization even if its us.

stevenj (changes requested):
We should not be allowing whole organizations.
In the case of Pallas this should not be needed as we should be upstreaming any changes we make, not maintaining them long term.

stevenj (approved):
LGTM

PR #164, opened 2024-01-18
fix: how CI CLI is handled in during setup

### LINKED ISSUE #161: 🛠️ [TASK] : Always build CI CLI when dogfooding
### Summary

Attempting to lock to a specific tag eventually breaks when a new version of the CLI that is not backwards compatible is released.

### Description

The [setup action](https://github.com/input-output-hk/catalyst-ci/blob/master/actions/setup/action.yml) should be modified as follows:

1. A conditional should be added to the "Install CI CLI" step that only runs the step when `cli_version` is _not_ `local`
2. A new step should be added at the end of the action steps (after Earthly has been configured):
   1. The step has a conditional to check if `cli_version` is set to `local` (i.e it only runs when it's local)
   2. The step uses the `run` action to call Earthly to build the `release` target to the CI CLI. It sets the `artifact` parameter to `"true"` and the `artifact_path` parameter to `/usr/local/bin`. 

After this, the `dogfood.yml` workflow should be modified to set the `ci_cli_version` parameter to `local`. The result should be that the CI always builds a local version of the CLI when running, thus avoiding a dependency on a release. By using Earthly (with a remote runner) the local build will be extremely efficient and shouldn't add too much time to the CI pipeline (TBD, we can re-evaluate if this is not true).

bkioshn:
# Description
Fixing how setup handles CI/CLI where the current setup action, which locks to a specific tag, faces issues with backward compatibility when a new, incompatible version of the CLI is released.

## Related Issue(s)
Closes #161 

## Description of Changes
- Remove the `local` flag in `install` - include removing the logic related to local in`install.ts` and install.test.ts`
- Modify `setup`  to handle `cli_version` where 
    - Case 1 `cli_version` is `local`
        - run the `release` target of the ci to build a artifact.
        - add another step to move the artifact to `/usr/local/bin` path
    - Case 2 `cli_version` is `<any other value>`
        - install the given version
- Readme fix
     - `actions/run.md`
     - `reference/actions.md`
     - fix unrelated topic in `style/index.md`


.github/workflows/ci.yml
```diff
@@ -196,7 +196,7 @@ jobs:
       earthly_runner_address: ${{ secrets.earthly_runner_address }}
       earthly_runner_secret: ${{ secrets.earthly_runner_secret }}
   deploy:
-    uses: input-output-hk/catalyst-ci/.github/workflows/deploy.yml@master
+    uses: input-output-hk/catalyst-ci/.github/workflows/deploy.yml@fix/ci-cli-build
```
minikin, line 199: Should it be here and below `master` instead of `fix/ci-cli-build`?

minikin (changes requested):

.github/workflows/ci.yml
bkioshn: Will change back to master before merge. This is for testing.

.github/workflows/publish.yml
```diff
@@ -88,13 +88,20 @@ jobs:
...
       - uses: actions/checkout@v3
       - name: Setup CI
-        uses: input-output-hk/catalyst-ci/actions/setup@master
+        uses: input-output-hk/catalyst-ci/actions/setup@fix/ci-cli-build
         with:
-          cli_version: ${{ inputs.ci_cli_version }}
           configure_registries: "false"
-          earthly_skip_install: "true"
+          aws_ecr_registry: ${{ inputs.aws_ecr_registry }}
+          aws_role_arn: ${{ inputs.aws_role_arn }}
```
Mr-Leshiy, line 95: didn't get why we need path these arguments as ` aws_ecr_registry`, ` aws_role_arn`, `aws_region` etc.   for the `setup` action, can you elaborate on this ?

Mr-Leshiy:
@bkioshn have you tested your fix ? I mean changed `cli` implementation, so as a result dogfood CI should fail

.github/workflows/publish.yml
bkioshn: If I'm not mistaken, it is required for running Earthly remote

bkioshn:
> @bkioshn have you tested your fix ? I mean changed `cli` implementation, so as a result dogfood CI should fail

@Mr-Leshiy The failed that we encounter last time only occurs when ci is changed and merge to master because of the ci version. Now I can only test that the `setup` really calling the release target when the version is set to local. However, will try changing the ci and see whether it works as I expected or not

stevenj (commented):
This PR looks OK to me.
@jmgilman Can you review for final approval, because its touching areas you have the most expertise with.

actions/run/src/run.ts
```diff
@@ -46,10 +47,12 @@ export async function run(): Promise<void> {
...
     core.info(`Running target: ${t}`)
     const argsSpawn = [...args]
-    // Artifact is set
-    if (artifact) {
-      core.info(`Pushing target ${t} with artifact tag`)
+
+    if (artifactSaveLocal) {
       argsSpawn.push('--artifact', `${t}/`, `${artifactPath}`)
+    } else if (artifactName.trim() !== '') {
+      core.info(`Pushing target ${t} with artifact tag`)
```
jmgilman, line 54: This is a confusing log entry. Pushing makes sense when you read the code, but looking through the logs it will be unclear what "pushing target" means (could easily be confused with pushing a container image).

actions/run/src/run.ts
```diff
@@ -3,7 +3,8 @@ import { spawn } from 'child_process'
 import * as path from 'path'
 
 export async function run(): Promise<void> {
-  const artifact = core.getBooleanInput('artifact')
+  const artifactSaveLocal = core.getBooleanInput('artifact_save_local')
```
jmgilman, line 6: I want to avoid this, if possible. These actions are used outside of CI in other places, so making backward incompatible changes will break them all. 

If I understand correctly, you want to explicitly set the artifact name instead of just dropping the artifacts into a folder. While this makes sense, I think it's unnecessary, because the artifact name is already set in the Earthfile. So, assuming you know the artifact name (which we do), you can use the action in the original form and just add an additional command afterwards:

```
mv ./artifacts/artifact_name /path/to/new/location
```

This would avoid making a breaking change.

actions/setup/action.yml
```diff
@@ -95,7 +101,18 @@ runs:
...
     if: ${{ inputs.earthly_runner_secret != '' && inputs.earthly_skip_install != 'true' }}
     with:
       secret: ${{ inputs.earthly_runner_secret }}
+  # CI CLI is set to 'local', run release to create ci artifact
+  # use buildkit-host tag
+  - name: Build CI CLI
+    uses: input-output-hk/catalyst-ci/actions/run@fix/ci-cli-build
+    if: ${{ inputs.cli_skip_install == 'false' && inputs.cli_version == 'local' }}
+    with:
+      artifact_name: 'ci'
```
jmgilman, line 114: In the original form, you should be able to set the `artifact` to a temp directory and then manually copy the resulting binary to `/usr/local/bin/` in a subsequent step.

jmgilman (changes requested):

actions/run/src/run.ts
bkioshn: Will change to  `Running --artifact with target ${t}, output to ${artifactPath}`

bkioshn: I changed to `Running --artifact with target ${t}, output to ${artifactPath}`

actions/run/src/run.ts
bkioshn: Yes, I think calling the artifact name directly would be much cleaner, but will revert and move the file instead.

Mr-Leshiy (dismissed):
LGTM

stevenj (dismissed):
I am good with this if @jmgilman is.

jmgilman (dismissed):
LGTM. Sorry for the delay (busy with voting launch) and thanks for working on this :)

PR #192, opened 2024-02-20
build(cargo deny): allow

cong-or:
Unable to pass cargo deny without the following. Problems came about after adding `cardano-follower` dependency.

We don't have versioning on the cardano-follower hence wildcard issue. Plus we need to whitelist hermes git repo.


cong-or:
@FelipeRosa is pallas still required here?

FelipeRosa:
> @FelipeRosa is pallas still required here?

It's not. I forgot to remove that.

cong-or:
ok, as discussed by changing the git allow; we will break other repos and will need to fix them accordingly to align and match them.

stevenj (approved):
LGTM

Issue #98, opened 2023-11-20
🛠️ [TASK] : Python linter


Mr-Leshiy:
### Summary

Add setup and example usage of Python linters

### Description

Need to add Python linter to the build process and add some common, standard configuration of it which would be used for any python project.
As a reference you can take a look how it was done for PostgreSQL builder https://github.com/input-output-hk/catalyst-ci/blob/master/earthly/postgresql/Earthfile

### Subtask list:
- [ ] Add linter
- [ ] Update Python guide

PR #193, opened 2024-02-22
feat: add python linter

alicechaitea:
# Description

Add linter for Python

## Related Issue(s)

Closes [#98](https://github.com/input-output-hk/catalyst-ci/issues/98)


bkioshn (changes requested):
The github action is not currently running the Python linter

earthly/python/Earthfile
```diff
@@ -22,23 +28,40 @@ python-base:
...
-
+    
     # Install Poetry to manage python.
     RUN curl -sSL https://install.python-poetry.org | python3 -
     # Prevent warning when running on machines with more than 6 cores.
     #   "Connection pool is full, discarding connection: pypi.org. Connection pool size: 10"
     RUN poetry config installer.max-workers 10
+    
+    # Install flake8 globally using Poetry
+    RUN pip3 install flake8
```
stevenj, line 39: We don't want to use flake8 as the linter.  We should be using `ruff` as its much much faster, includes flake8 checks and a ton of others all in the same tool.
See: https://github.com/astral-sh/ruff

Sorry, it should have been stated in the ticket.

earthly/python/Earthfile
```diff
@@ -22,23 +28,40 @@ python-base:
...
+    FROM +python-base
+
+    # Copy the source code directory to the container.
+    COPY $src_dir /app
+
+    WORKDIR /app
+
+    # Run flake8 to lint the Python code.
+    # Adjust flake8 options as needed.
+    RUN flake8 .
```
stevenj, line 67: ruff
```suggestion
    RUN ruff check .
```

examples/python/Earthfile
```diff
@@ -1,12 +1,67 @@
...
+    ENV PYTHONDONTWRITEBYTECODE=1
+    
+    # Install Poetry to manage python.
+    RUN curl -sSL https://install.python-poetry.org | python3 -
+    # Prevent warning when running on machines with more than 6 cores.
+    #   "Connection pool is full, discarding connection: pypi.org. Connection pool size: 10"
+    RUN poetry config installer.max-workers 10
+    
+    # Install flake8 globally using Poetry
+    RUN pip3 install flake8
```
stevenj, line 39: +builder should already have the linter installed, so no need to install it again.
We should always use +builder and not redefine the build environment because its not maintainable.

examples/python/Earthfile
```diff
@@ -1,12 +1,67 @@
...
+    FUNCTION
+    ARG opts=""
+
+    FROM +python-base
+
+    # Copy our dependencies.
+    COPY pyproject.toml poetry.lock .
+
+    # Install it all with poetry
+    RUN poetry install $opts
```
stevenj, line 51: This should be used from the python base earthfile above, and not redefined.

examples/python/Earthfile
```diff
@@ -1,12 +1,67 @@
...
+    FROM +python-base
+
+    # Copy the source code directory to the container.
+    COPY $src_dir /app
+
+    WORKDIR /app
+
+    # Run flake8 to lint the Python code.
+    # Adjust flake8 options as needed.
+    RUN flake8 .
```
stevenj, line 67: As should this.

stevenj (changes requested):
`ruff` has a config file which controls the linter.
We need a standard config in `earthly/python` for it, and that standard config should be validated against the ruff config in a python project.
Its an error if a python project being linted:

1. Does not have a `.ruff.toml` config file; and
2. Its not exactly the same as the standard one in `earthly/python`

There is code in the rust builder which checks if a controlled file matches its control version.  Look at that and copy how its done for this config file.

earthly/python/Earthfile
alicechaitea: Changed to ruff commit c90df2f16b81271f0563faeccfa1d25161ca46af

earthly/python/Earthfile
alicechaitea: Changed to ruff commit c90df2f16b81271f0563faeccfa1d25161ca46af

examples/python/Earthfile
alicechaitea: Not sure I've understood correctly, but I've tried addressing the duplicate installation in commit c90df2f16b81271f0563faeccfa1d25161ca46af

examples/python/Earthfile
alicechaitea: Not sure I've understood correctly, but I've tried addressing the duplicate installation in commit 27d54659b925602a0223f62f2799f9e70a34d69f

examples/python/Earthfile
alicechaitea: Took care of this in commit 27d54659b925602a0223f62f2799f9e70a34d69f as well

alicechaitea:
> The github action is not currently running the Python linter For reference you can see example from go

@bkioshn can you provide the exact location of the example from go?

earthly/python/Pillow-9.5.0-cp311-cp311-macosx_11_0_arm64.whl
stevenj, line 1: Why do we need this?  usually dependencies are built not installed from pre-built wheels.

examples/python/Pillow-9.5.0-cp311-cp311-macosx_11_0_arm64.whl
stevenj, line 1: Same question as above.

earthly/python/Earthfile
```diff
@@ -23,22 +27,30 @@ python-base:
...
+    FROM +python-base
+    # The setup steps for POETRY_SETUP would be included here, assuming it's defined elsewhere.
+    # This placeholder assumes the necessary Poetry and project setup is performed here.
+
 BUILDER:
     FUNCTION
-    ARG opts
-
-    # Copy our dependencies.
+    FROM +poetry
```
stevenj, line 47: Linting should not be part of the build, and this change to BUILDER will break all our python builds that use it.

There needs to be a `CHECK:` FUNCTION which uses the `builder`.
It needs to:
1. Validate the Ruff configuration in the project matches our fixed configuration (not done here yet).
2. Run ruff on the src

Thats all, it doesn't need to run poetry at all, and it shouldn't even assume we use poetry because we do have python code which needs linting but which is not a poetry package.  Such as our python based build scripts.

Then in those targets we can add a `check:` target which should just look like this:

```
check:
    COPY --dir ./src .
    DO ./../../earthly/python+CHECK
```

The CHECK function should look something like this (untested):
```
CHECK:
    FUNCTION

    FROM +python-base

    COPY .ruff.toml .

    RUN /scripts/std_checks.py
```

See: `examples/rust/Earthfile` for an example of using a `std_check.py` file to do the python checks (in this case its just validate the config file and run `ruff check .` but we could add more checks to this file in future.

You will need to make a `std_checks.py` and include it in the builder in `/scripts` so that it can be run by the CHECK function.

examples/python/Earthfile
```diff
@@ -1,12 +1,28 @@
...
+VERSION --global-cache --use-function-keyword 0.7
+
+# Assuming the detailed setup from the second snippet is defined in relevant Earthly targets.
 
 builder:
     FROM ./../../earthly/python+python-base
 
-    COPY --dir ./src .
-    DO ./../../earthly/python+BUILDER
+    # Copy the src directory and project-specific files into /poetry in the container.
```
stevenj, line 9: Don't replace functions that already exist with inline commands.
Otherwise every python target (of which there are many across all our repos) needs to be changed to remove the use of the function.

Also, as stated above,  `build:` doesn't need to lint the python code that should be done as a `check:` target before building even occurs.

earthly/python/Earthfile
```diff
@@ -23,22 +27,30 @@ python-base:
...
     RUN poetry config installer.max-workers 10
 
-# Common python builder setup.
-# Parameters:
-#   opts      : Options passed to the `poetry install` command.
+    # Install ruff for linting instead of flake8.
+    RUN pip3 install ruff
+
+# Utilize the POETRY_SETUP UDC for setting up Poetry within our environment.
+poetry:
```
stevenj, line 40: Why do we need this?  It doesn't add anything over `python-base` itself?

stevenj (changes requested):

bkioshn (commented):
Would be great to try adding some rules https://docs.astral.sh/ruff/linter/

earthly/python/Earthfile
bkioshn: Add on to what Steven said the `CHECK` function should be available for other repo to use too for example, this spelling check https://github.com/input-output-hk/catalyst-ci/blob/master/earthly/cspell/Earthfile is used in `cat-voice` https://github.com/input-output-hk/catalyst-voices/blob/main/Earthfile

earthly/python/poetry.lock
```diff
@@ -0,0 +1,153 @@
+# This file is automatically @generated by Poetry 1.7.1 and should not be changed by hand.
```
Mr-Leshiy, line 1: Why do we need poetry files here ?
All example project or project on which you gonna test the builder should be inside `examples` dir

earthly/python/pyproject.toml
```diff
@@ -0,0 +1,14 @@
+[tool.poetry]
```
Mr-Leshiy, line 1: same

earthly/python/Earthfile
alicechaitea: thank you! will look at them asap

earthly/python/Earthfile
alicechaitea: Deleted commit 3aeedb2559301de40cfccd457d699301e11235dc

examples/python/Pillow-9.5.0-cp311-cp311-macosx_11_0_arm64.whl
alicechaitea: Deleted .whl commit 8e68144397d1b24ae50eb3b50f569e63f2b06acf

earthly/python/poetry.lock
alicechaitea: Deleted the files in commit 4f22d7aea18d573a03bc3ca0b87efbc43add4da0

earthly/python/pyproject.toml
alicechaitea: Deleted the files in commit 4f22d7aea18d573a03bc3ca0b87efbc43add4da0

earthly/python/Earthfile
```diff
@@ -1,42 +1,105 @@
...
+    RUN ls -l /scripts && which python3 && echo $PATH
+
+    # Execute the check script
+    RUN /scripts/std_checks.py
+
+FORMAT:
+    FUNCTION
+
+    # Can not run format LOCALLY as it changes permissions of files to root:root
+    RUN --no-cache printf "%s\n%s\n%s\n" \
```
bkioshn, line 83: is this mean using `ruff format`?

examples/python/Earthfile
```diff
@@ -1,12 +1,35 @@
...
+    DO ./../../earthly/python+CHECK
+
+# format all Python files in the current project.  Local developers tool.
+#   CI target : false
+format:
+    LOCALLY
+
+    DO ./../../earthly/python+FORMAT --src=$(echo ${PWD})
+
+pythonlinter:
```
bkioshn, line 32: target name should be spereated with `-` like `python-linter`

examples/python/Earthfile
```diff
@@ -1,12 +1,35 @@
...
 
-    RUN poetry run pytest
+    DO ./../../earthly/python+CHECK
+
+# format all Python files in the current project.  Local developers tool.
+#   CI target : false
+format:
+    LOCALLY
+
+    DO ./../../earthly/python+FORMAT --src=$(echo ${PWD})
```
bkioshn, line 30: `FORMAT` doesn't accept any argument, if `ruff fmt .` is needed modify the `FORMAT` to use it. If not, this `format` target is not necessary

bkioshn (changes requested):
I think we need an update on the document too

earthly/python/Earthfile
alicechaitea: yes, automatically adjusting the code to adhere to ruff's linting protocols

examples/python/Earthfile
alicechaitea: yup I saw it in the CI error last night was planning to fix it this afternoon

examples/python/Earthfile
alicechaitea: can you check my latest commit 98485de21c7132e2bdb4c0c02ba4797d54783e5f

examples/python/Earthfile
alicechaitea: removed FORMAT commit 98485de21c7132e2bdb4c0c02ba4797d54783e5f

examples/python/Earthfile
alicechaitea: fixed in commit 98485de21c7132e2bdb4c0c02ba4797d54783e5f

earthly/python/Earthfile
alicechaitea: decided to delete FORMAT commit 98485de21c7132e2bdb4c0c02ba4797d54783e5f

earthly/docs/Earthfile
```diff
@@ -32,17 +32,13 @@ deps:
         tiff-dev \
         tk-dev \
         tcl-dev  \
-        git \
-        make
+        git
```
bkioshn, line 35: I don't think this file should be changed

earthly/postgresql/Earthfile
```diff
@@ -162,4 +162,3 @@ BUILD:
 
     # Push the container...
     SAVE IMAGE ${image_name}:latest
```
bkioshn, line 164: Add blank line

bkioshn (changes requested):

earthly/docs/Earthfile
alicechaitea: this is the original file when I've just branched out my PR
right now I'm trying to back track to why the CI is erroring

stevenj (dismissed):

stevenj (approved):
LGTM

PR #204, opened 2024-03-15
fix: add cc0-1 license to rust std config

bkioshn:
# Description

Add CC0-01 license in deny toml


stevenj (approved):
CC0-1.0 is a public domain declaration and is compatible with Apache 2.0

PR #229, opened 2024-04-10
feat(cat-ci): Add mithril targets that use official tools to get proper snapshot links

stevenj:
This PR adds a generalized Mithril snapshot downloader.
It supports both fully validated and unvalidated downloads.

Validated are slower, but are guaranteed not to have been tampered with.
We would need validated snapshots in production.

Due to the issues we have had with CI on snapshots and that preprod isn't very slow to download, preprod snapshots used
in CI are fully validated.  Other snapshots used for development purposes are not.

## Breaking Changes

This PR changes the name of the mithril target and so will break cat-voices integration test if it's not updated to match.


stevenj:
This PR is blocked pending an update to our cat-ci runner to support Earthly v0.8

earthly/mithril_snapshot/Earthfile
```diff
@@ -1,19 +1,86 @@
...
+        # This automatically handles unarchiving the snapshot
+        RUN mithril-client -vvv cardano-db download --json $snapshot_digest
+    ELSE
+        # This is faster, but it could be a faked snapshot file.
+        LET DLFILE=$(mithril-client cardano-db snapshot show --json $snapshot_digest | jq '.locations[0]' | tr -d '\"')
+        RUN mkdir -p ./db && \
+            wget -O - $DLFILE | zstdcat | tar -xf - -C ./db
+    END
+
+mainnet-snapshot:
```
Mr-Leshiy, line 45: WDYT to make all these targets which are produce artefacts as `package-*` ?
So it cat-ci we can see that they are produced correctly, otherwise ci for this repo will no trigger these targets

stevenj: Because i don't want mainnet to run in CI at all.
We can add a separate package target, which I already have to do it for preprod which we will use in CI.
The other targets are a developer convenience only at this point.

stevenj: @saibatizoku has a mithril tester which runs on preprod snapshot now, but it finishes in < 30 seconds when running in earthly.
I want to test against manner snapshot to see what happens over time, and that why I made this.

PR #232, opened 2024-04-15
fix: improve dependency installation in rust builder

### LINKED ISSUE #228: 🛠️ [TASK] : Enhance the rust builder to use techniques the same as the Earthly rust library
### Summary

Make Rust builds better at cache hits, and improve our Rust build times.

### Description

We are not hitting cache effectively with Rust builds.
And the base builders take a long to to create and are also not effectively being cached.
Updating a single tool can cause the whole rust builder to regenerate, not just the single tool.

The task here is to adapt:
https://github.com/earthly/lib/blob/main/rust/Earthfile
Into our rust builders so that we can get better caching.

We can not use the earthly rust library 100% as is, because we have separate requirements, but we can probably use INIT and adapt the other functions into our own rust ci setup.

One thing we can probably do better is tool building.
Currently all tools are installed with a sequence of RUN steps.
This is bad because one tool change causes all tools after it to rebuild, not just the one tool we want.

The idea for this is to create a `tools/Earthfile` which uses the base rust builder.
Targets in this file build a single tool, and save the tool as an artifact.

The final build with all tools can then just be the base rust tooling with each of these targets COPIED into it.
The advantage is they can:
* Build in parallel
* Share the build cache speeding up the tools that do build.
* A change to one tool only causes that tool to be rebuilt.
We may be able to use a similar model to this in other places but lets try it here first.

This should all be on top of the caching enhancements so that tool builds are faster and better cached rust builds.
The tool builds CAN use the earthly rust library AS IS.  We will only need to adapt it because we script cargo execution inside python scripts, and the earthly rust library doesn't support that.

bkioshn:
# Description

Improve `rust` related performance

## Related Issue(s)

Closes #228 

## Description of Changes
- Use `lib/rust` https://github.com/earthly/lib/tree/main/rust 
   - Additionally, FUNCTION `EXECUTE`, `CARGO`, and `COPY_OUTPUT` are modified and added to suit our usecase
- Use `IMPORT` instead of path when refering other Earthfile
- Rust tools parallel building
- Modifying the Rust document
 
Result of this changes:
- Changing of one tool doesn't effect the other tools to be rebuilt
- Efficient use of cache result in faster workflow
- Cleaner code implementation
 
## Related PR
- #236 
- #240 


earthly/rust/tools/Earthfile
```diff
@@ -0,0 +1,57 @@
+VERSION 0.8
+
+tool-cargo-nextest:
+    FROM ../+rust-base
+    RUN cargo install cargo-nextest --version=0.9.68 --locked
+    SAVE ARTIFACT /usr/local/cargo/bin/cargo-nextest cargo-nextest
```
apskhem, line 6: According to Earthly reference:

> If `<artifact-dest-path>` is not specified, it is inferred as `/`

So the below suggestion should also work, without specifying name twice.

```suggestion
    SAVE ARTIFACT /usr/local/cargo/bin/cargo-nextest
```

docs/src/guides/languages/rust.md
```diff
@@ -9,7 +9,7 @@ tags:
 # :simple-rust: Rust
 <!-- markdownlint-enable single-h1 -->
 
-<!-- cspell: words toolsets stdcfgs depgraph -->
+<!-- cspell: words toolsets stdcfgs depgraph Rustup -->
```
minikin, line 12: Can we move all these words to [project.dic](https://github.com/input-output-hk/catalyst-ci/pull/232/files#diff-c74e62111a947fe7699747e1193dd592b3346ab2267f7db4c78c994b2569cd49)?

minikin (changes requested):

earthly/rust/stdcfgs/rust-toolchain.toml
```diff
@@ -1,3 +1,3 @@
 [toolchain]
-channel = "1.75.0"
+channel = "1.77.2"
```
minikin, line 2: Shouldn't we use the latest stable version `1.78.0`?

minikin (changes requested):

earthly/rust/stdcfgs/rust-toolchain.toml
bkioshn: @stevenj Is there a reason why we are still using 1.77?

stevenj: We are now on 1.78 everywhere

earthly/rust/Earthfile
```diff
@@ -54,77 +50,232 @@ rust-base:
...
+#            If not defined, "coverage-report.info" will not be saved, but will be in the layer.
+#
+# This is heavily based off of the rust `CARGO` command in the Earthly library.
+# Updates to that library must be reflected in this function when updating the version of the earthly rust library.
+EXECUTE:
     FUNCTION
-    FROM +rust-base
+    DO rust+CHECK_INITED
+    ARG --required cmd
+    ARG args1
```
Mr-Leshiy, line 164: Maybe it will be better to compose all arguments and cmd on the clients side and remove all these `arg1`, `arg2` etc.
Because with such approach it could be a case when we need to add more, which does not seems like a best solution.
So on the client's side of the `EXECUTE` function it could be something like this.

```Earthfile
target:
   ...
   LET cmd = "foo"
   LET arg1 = "--arg1=arg1"
   LET arg2 = "--arg2=arg2"
   
   +EXECUTE --cmd="$cmd $arg1 $arg2"
```

utilities/dbviz/Earthfile
```diff
@@ -33,13 +35,16 @@ all-hosts-check:
 build:
     FROM +builder
  
-    RUN /scripts/std_build.py  --bins="dbviz/dbviz"
+    RUN echo cache breaker 1
```
Mr-Leshiy, line 38: Why do we need this ?

earthly/rust/Earthfile
stevenj: No, i prefer it the way I did it,  otherwise we push management of this Earthly quirk to all users.  (Ugly is subjective, but I think it is).

utilities/dbviz/Earthfile
stevenj: We don't its debug code that I missed removing.

.config/dictionaries/project.dic
```diff
@@ -70,7 +72,10 @@ rustdoc
 rustdocflags
 rustflags
 rustfmt
+Rustup
```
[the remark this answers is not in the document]
stevenj: Yup

Mr-Leshiy (approved):
LGTM

PR #240, opened 2024-05-05
feat(rust): Add parallel building of Tooling and Rust code

stevenj:
# Description

Adds parallel building of tools AND rust code


bkioshn (approved):
LGTM try running it and result in parallel

PR #262, opened 2024-06-12
fix: flutter ci

bkioshn:
# Description

- Disable flutter analytics
- Fix text color in installer
- Remove pre dependency installation in Chrome installation

### DIFF
--- a/earthly/flutter/installer/Earthfile
+++ b/earthly/flutter/installer/Earthfile
@@ -37,14 +37,13 @@ INSTALL_CHROME_LINUX64:
     ELSE 
         # Install Chromium
         RUN echo "\033[1;34mInstalling Chromium..."
-        RUN apt-get update
-        RUN apt-get install -y chromium
+        RUN apt-get update && apt-get install -y chromium
         RUN ln -s /usr/bin/chromium /usr/local/bin/google-chrome
         RUN google-chrome --version
 
         # Install Chromium-driver
         RUN echo "\033[1;34mInstalling Chromium-driver..."
-        RUN apt-get install chromium-driver
+        RUN apt-get update && apt-get install chromium-driver
         RUN chromedriver --version
     END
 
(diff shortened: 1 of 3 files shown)


earthly/flutter/installer/Earthfile
```diff
@@ -5,53 +5,44 @@ VERSION 0.8
...
+            gpg --dearmor | \
+            tee /usr/share/keyrings/google-chrome.gpg >> /dev/null
+        RUN echo deb [arch=amd64 signed-by=/usr/share/keyrings/google-chrome.gpg] https://dl.google.com/linux/chrome/deb/ stable main | \
+            tee /etc/apt/sources.list.d/google-chrome.list 
+        RUN apt-get update && apt-get install -y google-chrome-stable=${CHROME_VERSION}${PACKAGE_RELEASE}
+        RUN google-chrome --version 
 
         # Install Chromedriver
-        RUN echo "${BLUE}Installing Chromedriver..."
+        RUN echo "\033[1;34mInstalling Chromedriver..."
```
apskhem, line 29: Can you make a variable for each color? I think it's hard to interpret when embedding the color code directly on the message. It's also on several places.

minikin, line 29: Do we want to have `\033[1;34m` value here and below?

bkioshn: Try it already, sadly doesn't work 🥹

bkioshn: I saw you added color, so I didn't make any changes. 
Do you still want the color? @minikin

stevenj: I would say for `echo` we shouldn't use embeded color because it's terminal dependent.
If we need color output we should do it like so (untested):
```
    LET blue = $(tput setaf 4)
    LET normal = $(tput sgr0)
    RUN printf  "${blue}This text is blue${normal}"
```
Note: You have to set it back to normal, otherwise everything after that printf will also be blue, it doesn't turn off automatically on a newline.

bkioshn: Solved, should be easier to read now 👀
https://github.com/input-output-hk/catalyst-ci/actions/runs/9565415001/job/26368336240#step:5:1

earthly/flutter/installer/Earthfile
```diff
@@ -5,53 +5,45 @@ VERSION 0.8
...
 INSTALL_CHROME_LINUX64:
     FUNCTION
     ARG TARGETARCH
+    LET BLUE ='\033[0;34m'
 
     # https://googlechromelabs.github.io/chrome-for-testing/
     IF [ "$TARGETARCH" = "amd64" ]
-        LET DISTR = "Debian 12 (Bookworm)"
-        LET CHROME_VERSION = 125.0.6422.76
+        LET CHROME_VERSION = 126.0.6478.61
```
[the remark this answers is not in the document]
stevenj: It wont get the latest version, that violates the nature of the cache.  The cache has no way to know the version has changed.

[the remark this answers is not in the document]
kukkok3: Ah that's right! but if we make the version as a parameter I could run a test with `--no-cache` that retrives the latest version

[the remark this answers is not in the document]
bkioshn: @kukkok3 Will make the version modifiable by `ARG`

kukkok3 (approved):
LGTM

PR #271, opened 2024-06-24
chore: temporarily disables deployment step

jmgilman:
Recent refactorings in catalyst-world broke auto-deployments. It's on the backlog to fix. In the meantime, disable the deployment step so that CI stops showing up as red.


stevenj (approved):
OK

PR #304, opened 2024-09-06
feat: Enable LaTex mathematical equations format for MkDocs

Mr-Leshiy:
# Description

Allowing [LaTex](https://www.latex-project.org) mathematical equations formatting for MkDocs pages.
Inside `.md` files it is possible to use some of the LaTex entities to nicely represent math equations like this
```Latex
\begin{equation}
  f = \frac{1}{2}
\end{equation}
```
will be shown as 
Screenshot 2024-09-06 at 14 24 16

docs example [section](https://input-output-hk.github.io/catalyst-ci/branch/feat_docs_latex/appendix/examples/latex/)

### DIFF
--- a/docs/mkdocs.yml
+++ b/docs/mkdocs.yml
@@ -1,6 +1,6 @@
 INHERIT: std-theme.yml
 
-# cspell: words fontawesome
+# cspell: words fontawesome mathjax
 
 # Project Information
 site_name: Project Catalyst - CI Documentation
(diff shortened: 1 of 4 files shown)


kukkok3 (dismissed):
LGTM

stevenj (changes requested):
Looks fine, but can we add an example of its use in the: `docs/src/appendix/examples` pages.
This helps us not forget what features we have, and gives us a template to copy from when we want to use them.

It is good if it can be shown interacting with other features like admonitions.

So treat the example as a mini turorial on how to do it, and you can even include a link to what is the necessary format for the equations, because not everyone knows latex.

stevenj (approved):
LGTM

PR #306, opened 2024-09-11
feat(ci): Update CI to allow the chain-sync-v2 catalyst gateway branch to build cleanly.

stevenj:
# Description

As stated


earthly/rust/Earthfile
```diff
@@ -291,3 +291,16 @@ SETUP:
     # The ONLY toolchain we support is the one in the installer.
     # DO NOT copy the `rust-toolchain.toml` file into CI as it breaks
     # the toolchain on some targets and leads to build failure.
+
+
+# SYNC_STD_CFG : Syncs the standard config files locally.
+SYNC_STD_CFG:
```
Mr-Leshiy, line 297: What do you think to replace this target to `LOCALLY` target,
so the only thing will be needed is to define a `just`  target for a considered repo ?

bkioshn (approved):

earthly/rust/Earthfile
stevenj: Locally is always problematic with earthly.  Its why i started using just files.
I don;t know how LOCALLY could work here, because all this needs to do is save artifacts from a container into the src repo at the correct place, its not running anything locally.

PR #318, opened 2024-09-27
fix(cat-ci): Flutter broken CI cache

stevenj:
# Description

Adds a reliable version number to browser artifacts required by the flutter builder.

### DIFF
--- a/earthly/flutter/installer/Earthfile
+++ b/earthly/flutter/installer/Earthfile
@@ -1,8 +1,10 @@
 VERSION 0.8
 
+# cspell : words Rxxxx
+
 # DO NOT USE `main` branch.
 # When updating the artifacts, update them in `main` and then create a sequentially numbered branch from it called `Rxxxx`
-# Then replace that Realse version number in the path below.
+# Then replace that Release version number in the path below.
 # This ensures that the tools are properly cached, and will also be re-installed when the upstream artifact is updated.
 # If a branch is no longer used by CI, delete it.
 ARG --global BASE_URL = "https://github.com/input-output-hk/catalyst-storage/raw/R0001/web-browsers"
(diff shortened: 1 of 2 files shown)


earthly/flutter/installer/Earthfile
```diff
@@ -1,6 +1,13 @@
 VERSION 0.8
 
-ARG --global BASE_URL = "https://github.com/input-output-hk/catalyst-storage/raw/main/web-browsers"
+# cspell: words Rxxxx
+
+# DO NOT USE `main` branch.
+# When updating the artifacts, update them in `main` and then create a sequentially numbered branch from it called `Rxxxx`
+# Then replace that Release version number in the path below.
+# This ensures that the tools are properly cached, and will also be re-installed when the upstream artifact is updated.
+# If a branch is no longer used by CI, delete it.
```
dtscalac, line 9: Deleting "unused" branches might break older versions of CI that still depends on them.

If a branch is unused in the newest CI version but was used in previous CI version then after removing branch from catalyst-storage the previous CI version gets broken.

Shouldn't these branches stay alive as long as CI version that needs them stays alive? Do we remove CI versions or they stay forever?

stevenj: Thats what the last comment `If a branch is no longer used by CI, delete it1 means.
That we shouldn't delete the branch until we are sure its not in use by CI anymore.
If someone is on an old branch locally, they can easily update.

stevenj: I see what your saying.
In theory the CI versions stay forever, they aer just tags.
But we never use OLD Ci version for anything except "information".  And its not an external resource, so no one outside our group should be using cat-ci v1.3, and if they do, thats on them.

stevenj: In theory, old CI branches may already be broken.  Even upstream container images and tool versions disappear (or are yanked).  So there is no guarantee that any old (currently unused) ci version is still able to retrieve the assetts it relies on.

dtscalac: Alright, thanks for clarification

dtscalac (approved):
LGTM

Issue #315, opened 2024-09-25
🛠️ [TASK] : Integrate SchemaSpy into our Postgresql Docs building


stevenj:
### Summary

Update Postgresql Schema rendering to a better tool

### Description

SchemaSpy has much better schema rendering capabilites than we do currently.
See: https://schemaspy.org/

As we will need a very good way to render and review DB schemas, and their development, it is worthwhile improving how we visualize it.

The task here involves:

- [ ] Remove DB Viz from our Postgresql documentation process and replace it with SchemaSpy.
- [ ] Update the Example in the cat-ci docs to show how it is used.
- [ ] Remove the DBViz utility from cat-ci altogether (no longer required).
- [ ] Integrate the new schema documentation tooling into catalyst-voices, prior to working on DB schema updates there.

Note: 
* Catalyst-Core still uses dbviz, and there is no need to update it to use the new tooling.  Its version is pinned, so removing dbviz from a later version should not be a problem.
* No other repo currently has database schemas to document, so no requirement exists to update any other docs to this.

PR #331, opened 2024-10-14
refactor: update sql schema autogen tool

neil-iohk:
# Description

Replacement of dbviz, schema generation tool with SchemaSpy. Removal of old references to dbviz in the postgres earthly target and Catalyst-CI Postgres Example

Output Link For Docs:
https://input-output-hk.github.io/catalyst-ci/branch/chore_update_sql_schema_autogen_tool/appendix/examples/

## Related Issue(s)
https://github.com/input-output-hk/catalyst-ci/issues/315

## Description of Changes

- Modified the SQL schema autogen tool to removing existing dbviz tool and replacing with SchemaSpy for automated schema generation.
- Improved documentation around tooling,removing references no longer relevant to SchemaSpy autogen tool.
- Added markdown template and custom script to ensure the auto-generated SchemaSpy UI displays correctly and fills the screen removing double scroll bars in the view.
- Added Java Eartly stage to use as a base from which to copy necessary files during document build stage and improve caching.
- Refactored how Java dependencies are downloaded and cached. Added a base image openjdk:21-jdk-slim to avoid network calls like apt-get running on every build. This caches the Java installation and other dependencies to avoid unnecessary network calls during a build.


earthly/postgresql/Earthfile
```diff
@@ -43,10 +43,6 @@ postgres-base:
     # Get refinery
     COPY rust-tools+tool-refinery/refinery /bin
 
-    # Get dbviz
```
stevenj, line 46: Instead of trying to run another container, could you just copy the necessary tools for schemaspy into postgres-base here?

stevenj (changes requested):

earthly/postgresql/Earthfile
neil-iohk: packages not added in postgres-base

earthly/postgresql/Earthfile
```diff
@@ -34,8 +37,10 @@ postgres-base:
         ripgrep \
         wget \
         pipx \
-        && rm -rf /var/lib/apt/lists/*
-
+        && rm -rf /var/lib/apt/lists/* \
+        && wget -O /bin/schemaspy.jar https://github.com/schemaspy/schemaspy/releases/download/v${SCHEMASPY_VERSION}/schemaspy-${SCHEMASPY_VERSION}.jar \
```
stevenj, line 41: I would be inclined to put any `wget` in its own `RUN`.  Otherwise, a change in any of these versions will require all the dependencies being `apt-get` installed to be re-installed, when they are not needed to be re-installed.

If the JDBC_VERSION is in your opinion, less likely to change than the SCHEMASPY_VERSION it should also be `wgetted` first, so that its not necessary to pull it again if SCHEMASPY_VERSION changes, but it does not.

As soon as a RUN can not be cached, then every other command after it will no longer be cached.  This is one of the motivations of putting things in their own targets, to avoid unnecessary rebuilds/installs of dependencies.
So the ordering should be from "least likely to change"  to "most likely".  Some of this is subjective, so there is no strict "right" answer to the dependency install order, but it is something to consider.

By the same token, it might be worth moving the `java+COPY_DEPS` after the `apt-get` but before the wgets.

I would also consider putting the wgets after the sqlfluff install for the same reason, we change its version rarely.

stevenj (changes requested):

stevenj:
Also, because its no longer used, we can totally remove the `dbviz` tooling from this repo.  Add that to this PR.

stevenj (approved):

stevenj (approved):
LGTM

Issue #333, opened 2024-10-17
🛠️ [TASK] : Make github action/s which can pre populate and validate Project fields in PRs and Issues


stevenj:
### Summary

We want to make sure project fields are completed before tickets are close DONE, or PRs are merged.

### Description

The following are fields which we want completed for all Tickets closed as Done, and all PRs that are merged:

![image](https://github.com/user-attachments/assets/a5d5b373-057a-41ab-9435-a1b19b1da226)

In order to improve project tracking, it would be good to have the following automated:

1. Auto assign PRs to the author. See: https://github.com/marketplace/actions/auto-author-assign
2. Auto select the next milestone, when a PR or Issue is created. See: https://github.com/benelan/milestone-action
3. If possible, default Project Fields. to sensible defaults.  Priority Medium, and for a PR, set the iteration to the current one, and start date to today.
4. Before a PR can be merged, all fields need to be filled in.
5. Before a ticket is closed as DONE, all fields need to be filled in.

There are github actions which may help with some of this, or it may require custom actions to be written.
The aim is to streamline Ticket and PR creation and ensure all fields are reliably completed all the time.

This may require multiple actions.
The actions should be made in cat-ci, and referenced from our other repos to maintain uniformity of implementation and simplify maintenance.

PR #334, opened 2024-10-18
ci: Add project fields validator

minikin:
### Description 

- auto assign PR creator as  assignee
- verifies that all project fields are filled out and fail if any are left unfilled.

resolves [#333](https://github.com/input-output-hk/catalyst-ci/issues/333)


utilities/project-fields-validator/Earthfile
```diff
@@ -0,0 +1,32 @@
+VERSION 0.8
+
+IMPORT github.com/input-output-hk/catalyst-ci/earthly/python:v3.1.7 AS python-ci
+
+check:
```
stevenj, line 5:
```suggestion
test:
```
I suggest making this a `test` target, so that i won't run until after all the normal checks and builds have run.
Then its only going to block publishing/release or merging.

Mr-Leshiy (approved):
LGTM

minikin:
@stevenj, can you please review?
PS: The validate-project-fields job is failing because the GitHub token used in CI doesn't have sufficient permissions to access the project.

stevenj:
Will need to change the branch in the action to `main` so it can fetch the right code after the merge.

PR #344, opened 2024-10-31
feat(rust): Update `cargo deny` check and deny.toml config

Mr-Leshiy:
# Description

- Enabled cargo deny `RUSTSEC` vulnerability check
- Changed `deny.toml` configuration file check from strict to non-strict mode (means that the file could be different from the *golden* one, in places which are not specified by the *golden* config).
- Made `advisories.ignore` and `source.allow-git` fields flexible and configurable on the rust project side itself and not strictly defined by `cat-ci` *golden* `deny.toml` config file

### DIFF
--- a/earthly/rust/scripts/std_checks.py
+++ b/earthly/rust/scripts/std_checks.py
@@ -90,7 +90,7 @@ def main():
     results.add(
         vendor_files_check.toml_diff_check("/stdcfgs/clippy.toml", "clippy.toml")
     )
-    results.add(vendor_files_check.toml_diff_check("/stdcfgs/deny.toml", "deny.toml"))
+    results.add(vendor_files_check.toml_diff_check("/stdcfgs/deny.toml", "deny.toml", strict=False))
 
     # Check if the rust src is properly formatted.
     res = exec_manager.cli_run("cargo +nightly fmtchk ", name="Rust Code Format Check")
@@ -104,7 +104,7 @@ def main():
     results.add(exec_manager.cli_run("cargo machete", name="Unused Dependencies Check"))
     # Check if we have any supply chain issues with dependencies.
     results.add(
-        exec_manager.cli_run("cargo deny check --exclude-dev -W vulnerability", name="Supply Chain Issues Check")
+        exec_manager.cli_run("cargo deny check --exclude-dev", name="Supply Chain Issues Check")
     )
 
     results.print()
--- a/earthly/docs/dev/local.py
+++ b/earthly/docs/dev/local.py
@@ -90,10 +90,10 @@ class DocsContainer:
             DocContainerNotFoundError: If the container with the specified name is not found.
         """
         # Make sure the container target is up-to-date
-        # try:
-        #     run_command(["earthly", self.target])
-        # except ProcessRunError as exc:
-        #     raise DocBuildError from exc
+        try:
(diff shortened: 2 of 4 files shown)


earthly/rust/scripts/std_checks.py
```diff
@@ -90,7 +90,7 @@ def main():
     results.add(
         vendor_files_check.toml_diff_check("/stdcfgs/clippy.toml", "clippy.toml")
     )
-    results.add(vendor_files_check.toml_diff_check("/stdcfgs/deny.toml", "deny.toml"))
+    results.add(vendor_files_check.toml_diff_check("/stdcfgs/deny.toml", "deny.toml", strict=False))
```
stevenj, line 93: No, i want strict checking of cargo-deny config.  Projects should not be able to modify it to suit themselves.

earthly/rust/scripts/std_checks.py
```diff
@@ -104,7 +104,7 @@ def main():
     results.add(exec_manager.cli_run("cargo machete", name="Unused Dependencies Check"))
     # Check if we have any supply chain issues with dependencies.
     results.add(
-        exec_manager.cli_run("cargo deny check --exclude-dev -W vulnerability", name="Supply Chain Issues Check")
+        exec_manager.cli_run("cargo deny check --exclude-dev", name="Supply Chain Issues Check")
```
stevenj, line 107: All the reasons for ignoreing the RUST-SEC vulnerabilites as errors remain,
A single change to the rustsec database can make our code broken and give us no viable way to resolve it.
I am not filling the cargo-deny config with tons of "We can't fix this vulnerability" exclusions.  Its just wrong.
If rust let us actually have some meaningful way to fix vulnerabiliteis, that would be one thing, but its just a timebomb otherwise waiting to break CI outside of our control.

stevenj (changes requested):

stevenj:
I would need to know a compelling reason why we must do this before I could approve it?

Mr-Leshiy:
@stevenj 
My general idea here is to enable RUST-SEC vulnerability detection and make `advisories. ignore` (which is actually a RUST-SEC ignores) and `allow-git` fields configurable for project.
I dont have a strong position about enabling RUST-SEC vulnerability detection, it just looks not problem if we will have a possibility to easily add exceptions. From what we can see now, we need to have about 1-5 exceptions per Rust project.

With such change it slightly reduces the headache to always modify in all our rust projects these fields if it is an issue only for **one** repository. So with this patch it will be possible just to add an exception for particular rust package.
 
 Also we are not actually loosing quality in general, because with this non-strict check other fields which are present in the golden config `earthly/rust/stdcfgs/deny.toml` are strictly verifiable (like license configs etc.), non-strict check just dont care about fields which are **not present** inside the golden config.

stevenj:
We don't use a lot of git projects, and they are already listed.  And I already said I don't want us to continually have to add ignores to the RUST-SEC section, and with this cahnge across multiple repos. its pointless if we have no way of dealing with them properly.

This change could require us to add ignores to multiple projects/repos because of a single RUST-SEC vulnerability that emerges.  And we still have no way of properly correcting them, because rust doesn't allow us to force transitive dependencies to a fixed version..

Mr-Leshiy:
@stevenj Got it.
So I will close this PR then.

PR #356, opened 2024-12-02
fix(flutter): flutter apt-get install & update as single command

damian-molinski:
# Description

Making sure that both `apt-get install` and `apt-get update` are called together so if either fails, the whole step fails and is not cached.

+ additional cleaning afterwords so smaller image size.


kukkok3 (approved):
LGTM

minikin (approved):

PR #383, opened 2025-02-14
fix: explicit https flutter fetch

damian-molinski:
# Description

Explicitly sets set arm64 flutter source as HTTPS because otherwise docker may attempt using SSH


dtscalac (approved):
LGTM

PR #384, opened 2025-02-20
Update Rust version and revision

stanislav-tkach:
# Description

- Update Rust revision to 2024.
- Update Rust version to 1.85.
- Enable the Clippy's `allow-indexing-slicing-in-tests` options.
- Add the `cardano_blockchain_types::Slot` type to the list of exceptions for the `arithmetic-side-effects` lint.

* [ ] My code follows the style guidelines of this project
* [ ] I have performed a self-review of my code
* [ ] My changes generate no new warnings
* [ ] New and existing unit tests pass locally with my changes

### DIFF
--- a/examples/rust/Cargo.toml
+++ b/examples/rust/Cargo.toml
@@ -8,7 +8,7 @@ members = [
 [workspace.package]
 version = "0.0.1"
 license = "MIT OR Apache-2.0"
-edition = "2021"
+edition = "2024"
 
 [workspace.lints.rust]
 warnings = "deny"
--- a/examples/rust/clippy.toml
+++ b/examples/rust/clippy.toml
@@ -1,4 +1,5 @@
 allow-unwrap-in-tests = true
 allow-expect-in-tests = true
 allow-panic-in-tests = true
-arithmetic-side-effects-allowed = ["num_bigint::BigInt"]
+allow-indexing-slicing-in-tests = true
+arithmetic-side-effects-allowed = ["num_bigint::BigInt", "cardano_blockchain_types::Slot"]
--- a/examples/rust/rust-toolchain.toml
+++ b/examples/rust/rust-toolchain.toml
@@ -1,3 +1,3 @@
 [toolchain]
-channel = "1.83"
+channel = "1.85"
 profile = "default"


stanislav-tkach:
it seems that the official 1.85 docker image isn't published yet.

earthly/rust/stdcfgs/clippy.toml
```diff
@@ -1,4 +1,5 @@
 allow-unwrap-in-tests = true
 allow-expect-in-tests = true
 allow-panic-in-tests = true
-arithmetic-side-effects-allowed = ["num_bigint::BigInt"]
+allow-indexing-slicing-in-tests = true
+arithmetic-side-effects-allowed = ["num_bigint::BigInt", "cardano_blockchain_types::Slot"]
```
stevenj, line 5: Is it possible to just disable the lint on the code with an annotation, rather than add the exception here.  I would rather we only use this list for code we don't control.

stevenj (changes requested):

earthly/rust/stdcfgs/clippy.toml
stanislav-tkach: Yes, it is possible to use annotations and it is what we currently do, but it is also possible to do that for `BigInt`. Personally I think that it is annoying to put annotations in multiple places even if we know that it is safe. Currently we have multiple annotations like this one:
```rust
// It is ok because slot implement saturating subtraction.
#[allow(clippy::arithmetic_side_effects)]
let slot_rollback_size = head_slot - rollback_slot;
```

stanislav-tkach: Anyway, I have removed `Slot` from this list. 🙃

stevenj (changes requested):
No further work on this PR is to be conducted until its been prioritized through the proper processes.

earthly/rust/stdcfgs/clippy.toml
stevenj: Putting it in the list means there is no way to know what code is exactly causing the issue, its globally disabling it.  Putting the annotation means this EXACT line is causing this issue and allows it to be properly validated.

In the above example thats an indication the API isn't great for that type.
instead it might be better:
```
let slot_rollback_size = head_slot.rollback_size(rollback_slot);
```
no need for the annotation.

Not suggesting this change be made, just that there are different ways to approach this.  I would much rather we be explicit so we can properly validate that the annotation is in fact safe and doing what we would expect, rather than silently squashing all instances of it and hoping one of those is not a problem.

